PR #8, opened 2020-01-07
vault-self-signed-support - adds support for vault using a self signe…

Alevsk:
…d certificate by adding the insecure configuration

### DIFF
--- a/cmd/kes/config.go
+++ b/cmd/kes/config.go
@@ -51,6 +51,7 @@ type serverConfig struct {
 			Addr      string `toml:"address" yaml:"address"`
 			Name      string `toml:"name" yaml:"name"`
 			Namespace string `toml:"namespace" yaml:"namespace"`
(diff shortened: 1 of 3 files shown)


harshavardhana:
We shouldn't be using insecure mode, is there a good reason to provide this option?

Alevsk:
> We shouldn't be using insecure mode, is there a good reason to provide this option?

We need to use kes and vault with a self signed certificate for testing in our local environments, I'll talk with @aead to see if there's an alternative on how to do this.

aead:
@Alevsk Closing this b/c a) already supported via https://github.com/minio/kes/blob/master/server-config.toml#L123 and b) we should never skip the X.509 certificate trust chain checks. Either disable TLS  (local testing) or role out PKI/direct trust in production.

Alevsk:
Thanks for the clarification @aead

PR #11, opened 2020-01-25
split /v1/key/create API into create and import

aead:
This commit splits the `/v1/key/create` API into
two separate APIs:
 - `/v1/key/create`
 - `/v1/key/import`

The reason for this API separation is that some
applications may be allowed to create keys - e.g.
a new secret key on their first startup. However,
this applications should probably not be allowed
to specify the secret key itself.

An application that can specify the secret key itself
may be a security issue since it does not necessarily
need the KES server to perform cryptographic operations
with this key.

Therefore, applications should only have the `v1/key/create`
permission (if required) but not the `/v1/key/import` permission.

However, the `/v1/key/import` API may be useful when migrating
from one backend store to another (e.g. from Vault to AWS and vice
versa).

**This is an API-breaking change** - However:
 - We have not committed to a stable API, yet.
 - Most applications that use the `/v1/key/create`
   API will not send any request body (secret key)
   since they don't want to create the key themself.
 - I'm not aware of any client using this API in a way
   such that this change would break them at the moment.
 - The CLI interface will handle the server API separation
   transparently.

### DIFF
--- a/client.go
+++ b/client.go
@@ -46,7 +46,30 @@ func NewClient(addr string, config *tls.Config) *Client {
 
 func (c *Client) Transport(transport http.RoundTripper) { c.httpClient.Transport = transport }
 
-func (c *Client) CreateKey(name string, key []byte) error {
+// CreateKey tries to create a new master key with
+// the specified name. The master key will be generated
+// by the KES server.
+func (c *Client) CreateKey(name string) error {
+	url := fmt.Sprintf("%s/v1/key/create/%s", c.addr, name)
+	req, err := http.NewRequest(http.MethodPost, url, nil)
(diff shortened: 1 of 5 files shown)


aead:
@Alevsk According to https://github.com/minio/m3/blob/6432e825d0ce9a2a58a6b7c92dcb7c2b985332af/cluster/kes.go#L179 this change will not affect minio/m3 - Can you confirm?!

Alevsk:
@aead I finished testing this PR with the current version of mkube and it's working fine 👍, please go ahead.

aead:
Thanks for testing @Alevsk

PR #16, opened 2020-02-11
stabilize server error API

aead:
This commit stabilizes and unifies the error interface
of the KES server. From now on the KES server returns
a error code >= 400 (bad request) and a JSON error message
as response body.

The error response has the following (simple) form:
```
{
  "message": "...",
}
```

Now, a client can rely upon the server returning an error status
code and a JSON object explaining why the error occurred.
Further, the client can now easily check whether the server
returned an expected error - e.g. `key not found`

However, this change does not finalize the errors returned by the server
themself but just the general structure of error messages.


aead:
> We can use xerrors instead of errors as import everywhere for consistency

Sure, done.

aead:
PTAL @harshavardhana

aead:
Closing this in favor of general project structure improvements PR.

PR #19, opened 2020-02-18
restructure package API

aead:
This commit restructures the packages of the kes project
such that:
 - The root `kes` (`github.com/minio/kes`) contains the
   client API and any types needed for that. E.g. `Policy`.
 - The `cmd/kes` package contains the client CLI and server
   application code.
 - All packages under `internal` (e.g. `internal/auth`) contain
   functionality for the server application and cannot be imported
   by external packages.

This has the following implications:
 - Everything in the root `kes` is part of the public API and has
   to be sem-ver compatible.
 - The root `kes` **must not** import anything in `internal/*`
 - Anything in `cmd/kes` or in `internal/*` **may** import
   anything from `kes`.

In general, all packages under `internal` should be self-contained
and domain-specific. The applications in `cmd/kes` use these packages
as building blocks for their application logic.

Note: This commit only restructures the packages but does not refactor
them (completely) to limit the scope of this commit.


aead:
@harshavardhana PR is in service of #16 - Functionality of #16 will be added on top of this, as separate PR.

PR #25, opened 2020-03-03
add initial KMS support for sealed secrets

aead:
This commit implements encrypted secrets.
Before, KES stored all secrets as plaintext at the
backend secret store. Therefore, KES assumed/required
that the secret store is a secure/trusted component.

While this assumption may be true for secret stores like
Vault or AWS SecretsManager it is not true in the general
case. Further, even if the secret store is considered trusted
it may be desirable to separate the trust between the secret
storage provider and another party.

Therefore, a KES server may want to encrypt secrets before
storing them at the secret store. In general, KES encrypts
secrets with an (external) KMS - like Vault, AWS-KMS or
Azure Vault. Therefore, KES sends a secret to the KMS before
writing it to the secret store. However, it doesn't send the
secret as plaintext to the KMS since that would reveal the
secret to the operator of the KMS / anyone who can inspect
KMS operations.

Instead, it generates a random (local) cryptographic key
when creating a secret and encrypts the secret with that key
*before* sending the secret to the KMS. So, the KMS only
sees ciphertexts (`Enc(local_key, secret)`). The local key
is then embedded into the value written to the secret store.

This construction ensures that neither the secret store provider
nor the KMS provider can decrypt an encrypted secret on their own.
So, it is possible to leverage e.g. a cloud KMS provider (AWS, Google,
MS, ...) to encrypt secrets on some local storage without sharing
these secrets with the cloud provider.

***

Therefore, this commit changes the architecture of the exiting code
quite a bit:

There is now a local store (`secret.Store`) and a remote store
(`secret.Remote`). The `secret.Store` implements en/decryption
(via the `secret.KMS` interface) and is responsible for caching
secrets.
The `secret.Remote` interface defines what operations a secret
store has to provide and must be implemented by all secret store
implementations.

Since, a secret store now can contain plaintext and encrypted
secrets the `secret.Remote` consumes strings while `secret.Store`
consumes `secret.Secret`.
However, for now we demand that the secret store returns only
plaintext secrets or encrypted secrets - but not both. When
the KES server is configured to use a KMS it will always try
to decrypt values from the secret store - including plaintext
secrets which should fail.

An encrypted secret has the following format:
```
{
  "local_key": {
    "bytes": "4BIYQkplmutXWkMg38enraePnsYQauIQkhzBhD7xs+U="
  },
  "kms_key": "5f91e7d9-a376-47cd-ad6f-3485abf9122c",
  "ciphertext":"<base64-encoded-ciphertext>"
}
```

The `local_key` is the cryptographic key used to encrypt
the plaintext secret before sending it to the KMS.
The `kms_key` is the name of the cryptographic key at
the KMS used encrypt the secret (ciphertext).
So, the result of encrypting the plaintext first with
the local key and then with the KMS is `ciphertext`.

***

Apart from this conceptual change, this commit also
adds a `secret.KMS` implementation for the AWS-KMS and
extends the KES server configuration to provide an
AWS-KMS configuration.

### DIFF
--- a/cmd/kes/config.go
+++ b/cmd/kes/config.go
@@ -51,6 +51,21 @@ type serverConfig struct {
 		} `toml:"audit" yaml:"audit"`
 	} `toml:"log" yaml:"log"`
 
+	KMS struct {
+		AWS struct {
+			Addr   string `toml:"address" yaml:"address"`
+			Region string `toml:"region" yaml:"region"`
+
+			Key string `toml:"key" yaml:"key"`
+
+			Login struct {
+				AccessKey    string `toml:"access_key" yaml:"access_key"`
+				SecretKey    string `toml:"secret_key" yaml:"secret_key"`
+				SessionToken string `toml:"session_token" yaml:"session_token"`
+			} `toml:"credentials" yaml:"credentials"`
+		} `toml:"aws" yaml:"aws"`
+	} `toml:"kms" yaml:"kms"`
+
 	KeyStore struct {
 		Fs struct {
 			Dir string `toml:"path" yaml:"path"`
--- a/cmd/kes/server.go
+++ b/cmd/kes/server.go
@@ -172,7 +172,25 @@ func server(args []string) error {
 		}
 	}
 
-	var store secret.Store
+	var store = &secret.Store{}
+	switch {
+	case config.KMS.AWS.Addr != "":
+		kms := &aws.KMS{
+			Addr:   config.KMS.AWS.Addr,
+			Region: config.KMS.AWS.Region,
+			Login: aws.Credentials{
+				AccessKey:    config.KMS.AWS.Login.AccessKey,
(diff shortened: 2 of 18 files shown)


aead:
> Quick question @aead is this backward compatible? - is sealing secrets opt in?

Yes, the KMS configuration for sealing secrets before writing them to the key store is optional. It allows you to turn e.g. the (insecure) filesystem backend into a "secure" one by encrypting the secrets with the external KMS. The same is the case for a potential etcd key store backend.
Further you can separate trust when using e.g. AWS and Vault. So you can use Vault to encrypt secrets before uploading to AWS and vice versa.

aead:
Please do not merge this PR, yet. There is an additional change I want to make that prevents the KMS provider learning the encrypted secrets.

aead:
Again, sry please do not merge. There is one **important fix** (not directly related to this PR) I have to make before this PR can be merged.

PR #27, opened 2020-03-04
ignore empty static credentials

### LINKED ISSUE #26: aws kms does not use ec2 instance role
#### Bug description

Create a configuration for AWS KMS that does not specify an access_key, secret_key, or session_token.

#### Expected behavior

KES should use the ec2-instance role's credentials.  The AWS SDK  supports this if you use th default provider.

aead:
This commit fixes a bug affecting the AWS SecretsManager
key store.
When no access key, secret key or session token is provided
the credentials passed to the AWS SDK should be `nil`. This
causes the AWS SDK to look for credentials in AWS env. variables,
or EC2 metadata.

I've verified that this commit fixes the bug reported in #26 by
creating an AWS EC2 instance and specifying no credentials in the
config file.
If the correct role (e.g. SecretsManager Read/Write) is attached to
the instance specifying no credentials causes the kes server (via AWS
SDK) to fetch the temp. credentials from the EC2 metadata.

Fixes #26

### DIFF
--- a/internal/aws/secrets-manager.go
+++ b/internal/aws/secrets-manager.go
@@ -201,15 +201,29 @@ func (store *SecretsManager) Delete(name string) error {
 // Authenticate tries to establish a connection to
 // the AWS Secrets Manager using the login credentials.
 func (store *SecretsManager) Authenticate() error {
+	credentials := credentials.NewStaticCredentials(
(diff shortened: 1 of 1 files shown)


harshavardhana:
Ideally this should be implemented in aws-sdk @aead

PR #32, opened 2020-03-23
aws-kms: honor error log configuration

aead:
This commit fixes a bug when initializing AWS-KMS.
Before, the error log was not passed to the AWS-KMS
client such that if the client would log a error message
it would always be written to the STDOUT/STDERR - regardless
what was specified in the config file.

However, this bug does not affect any (official) release
because it has been introduced by 817a824.


aead:
ping @harshavardhana

PR #35, opened 2020-03-27
add new `log` cmd and deprecate `audit` cmd

aead:
This adds a new CLI sub-command `log` that
provides mechanisms to work with KES server
logs - e.g. trace the server log event stream.

In general, there are two types of log events
produced by a KES server:
 - audit events
 - error events

The new `log` command will be the starting point
for dealing with both event types. At the moment
the CLI only deals audit event streams. Therefore,
the new `log` command preserves the exact behavior
of `audit`.
So, there is a sub-command `trace` and
`kes log trace` behaves exactly as `kes audit trace`
did before.

In the future, we want to add more and more log
management features to `kes log` - e.g. support
for error log events.

### DIFF
--- a/cmd/kes/log.go
+++ b/cmd/kes/log.go
@@ -1,3 +1,7 @@
+// Copyright 2020 - MinIO, Inc. All rights reserved.
+// Use of this source code is governed by the AGPLv3
+// license that can be found in the LICENSE file.
+
 package main
 
 import (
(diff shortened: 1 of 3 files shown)


aead:
This PR depends on #33 - therefore, do-not-merge.

Alevsk:
Looks good @aead

PR #39, opened 2020-04-13
add support for etcd (v3) as key store

aead:
This commit adds support for etcd as key store.
In general, the etcd configuration just requires
one/multiple server/cluster endpoints:

```
[keystore.etcd.v3]
address = [ "http://127.0.0.1:2379" ]
```

However, in production the etcd cluster
should expose a TLS/HTTPS endpoint and
the KES server should authenticate itself
via username/password or mTLS.

For username/password authentication use:
```
[keystore.etcd.v3.credentials]
username = ""
password = ""
```

For mTLS authentication use:
```
[keystore.etcd.v3.tls]
key  = "" # Path to private key
cert = "" # Path to certificate
```

The configuration uses the `etcd.v3` schema since
there is also a (quite old) v2 and there will be a
v4 eventually. By separating the configuration for
different versions we can keep the configuration
options of each version clean and minimal w.r.t. to
the etcd capabilities.

An alternative would have been to use a configuration
entry:
```
[keystore.etcd]
version = "v3"
```
However, if etcd makes some fundamental changes then
the configuration between versions eventually differs
significantly.
Basically, newer etcd versions could make our etcd
configuration messy in this case.
Therefore, this approach has been discarded.

The etcd client SDK (v3) relies on a third-party logging
framework (https://pkg.go.dev/go.uber.org/zap) instead of
the standard library.
This is fine, in general. However, etcd only allows logging
configuration via a `zap.Config` - which does not accept a
custom logging sink as `io.Writer`. This is a fundamental problem
since the KES server allows error log tracing via a client
and may support writing error logs to a web-hook.

This is not possible with the current combination of `zap`
and ectd client SDK. To avoid being inconsistent w.r.t. logging
errors (e.g. error log tracing just prints a subset of the log
events) we disable the internal etcd SDK logging completely.

While this is unfortunate there is not much we can do about this
at the moment.

To test this feature you may:
 1. Start a etcd server: `etcd`
 2. Set the etcd endpoint in the KES server config to:
    ```
    [keystore.etcd.v3]
    address = [ "http://127.0.0.1:2379" ]
    ```

Further testing requires adding more nodes and/or enabling
authentication mechanisms. Refer to: https://etcd.io/docs/


aead:
This PR requires some testing (esp. multi-node, user roles and mTLS).

aead:
Will also follow up on the reported licensing issue.

internal/etcd/etcd.go
```diff
@@ -0,0 +1,320 @@
...
+
+		// Etcd does not allow us to specify a standard
+		// library logger nor configure the log sink as
+		// io.Writer (only paths / URLs). Therefore, we
+		// disable the etcd-internal logging by using
+		// the special "/dev/null" path of the zap package.
+		// It instructs zap to not write logs.
+		//
+		// This entire situation is unfortunate but there
+		// is not much we can do here. ¯\_(ツ)_/¯
```
Alevsk, line 135: ¯\_(ツ)_/¯

aead:
The FOSSA check now passes and the direct dep. seem to be AGPLv3 compatible. Therefore, previous FOSSA test failures seemed to be a bug. Therefore, consider the licensing issue as resolved.

PR #42, opened 2020-04-17
restrict server API to HTTP2

aead:
This commit restricts the KES server API to
HTTP2.

The reason for this restriction is that future
API functionality probably will require bidirectional
streams between the server and the client.

In particular, the KES server will expose APIs that
return a response to the client even though the client
still sends more data. (streaming)

This is not possible with HTTP/1.0 and HTTP/1.1 since
the server must not return a response body to the client
before the server hasn't read the entire client request.

While, in theory, it would be possible to restrict only
those APIs to HTTP2, it seems better to enforce this
restriction for all APIs from the beginning. This avoids
mental complexity for clients - since they don't have to
remember which APIs are restricted and which are not.

Also, we can relax the restriction in the future much more
easily then putting additional limitations in place. The
later would break clients.


aead:
> Why can't you turn off http.Server.Protos to http2 only ?

AFAICS this does not work. Neither the [`NextProtos`](https://golang.org/pkg/crypto/tls/#Config) nor the [`TLSNextProto`](https://golang.org/pkg/net/http/#Server) can be used to disable HTTP/1.0 or HTTP/1.1. AFAICS with `NextProtos` you can only specify the order preference for ALNP while the purpose of `TLSNextProto` seems to be an update-mechanism for the http.Server config and the TLS connection.

For example the following diff does not prevent clients from using HTTP/1.0:
```
diff --git a/cmd/kes/server.go b/cmd/kes/server.go
index 9b8d140..2ba826c 100644
--- a/cmd/kes/server.go
+++ b/cmd/kes/server.go
@@ -304,6 +304,7 @@ func server(args []string) error {
                Handler: mux,
                TLSConfig: &tls.Config{
                        MinVersion: tls.VersionTLS13,
+                       NextProtos: []string{"h2"},
                },
                ErrorLog: errorLog.Log(),
```
Can be verified via `curl --key <key> --cert <cert> https://127.0.0.1:7373 -k --http1.0`

aead:
@harshavardhana Have I addressed your [comment](https://github.com/minio/kes/pull/42#issuecomment-617186716) or did I miss something?

PR #43, opened 2020-04-28
refactor server CLI and configuration

aead:
This commit simplifies the KES server CLI
and configuration based on an internal
design review.

From now on, the KES server only accepts
YAML configuration since the most common
why to use/deploy a KES server is as (docker)
container resp. as K8S pod.
Some configuration fields have changed as well.
For example the previous Vault key store `name`
field has been replaced by the `prefix` field.

Further, this commit simplifies the server
CLI. The TLS private key can be specified
via `--key` and the certificate via `--cert`
(Same as for curl).

Further, the authentication can now be disabled
via `--auth=off` or explicitly enabled via
`--auth=on`. As before mTLS authentication
verification is enabled by default.
(`--auth=on` is the default)


Alevsk:
toml configuration is not longer supported then?

aead:
> toml configuration is not longer supported then?

Yeah, I like toml but yaml is more K8S-friendly and since we make breaking changes and simplify then let's make all the breaking changes now.

PR #56, opened 2020-06-01
vault: allow custom engine paths

### LINKED ISSUE #55: Unable to use vault when secret backend is in non-standard location
The Vault provider seems to add hardcoded "/kv" prefix to the supplied location:

https://github.com/minio/kes/blob/5de0d193377e0373cf769b6daec7ed316bbba191/internal/vault/vault.go#L159

This is a problem when we have more than one KV secret engine mounted on Vault or one choose to use a non-standard path.

I think the easiest thing to do here would be to simply not add the prefix, leaving to the user the liberty to specify the location. Another option would be to add another option to specify this prefix, with "kv" being its default value.

aead:
This commit allows users to specify a custom
engine path for the K/V secret engine and
AppRole authentication method, and therefore,
fixes #55.

Vault allows engines to be mounted at arbitrary
paths. For instance, you can mount the K/V secret
engine at `/kv` or `/secrets`.
Further, Vault allows multiple instances of the same
engine to be present at the same time. Therefore,
supporting user-defined engine paths is necessary.

The same is true for authentication methods/engines,
like AppRole.

This adds two configuration options `engine: <engine-name>`.
One is for the K/V engine and the other one is for the
AppRole engine.
However, if not specified the server assumes that the
K/V engine is mounted at the "default" `/kv` path and
similarly, that the AppRole engine is mounted at the
"default" `/approle` path.

### DIFF
--- a/internal/vault/vault.go
+++ b/internal/vault/vault.go
@@ -19,6 +19,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"path"
 	"time"
 
 	vaultapi "github.com/hashicorp/vault/api"
@@ -31,6 +32,7 @@ import (
 // authentication should be retried
 // whenever it fails.
 type AppRole struct {
+	Engine string // The AppRole engine path
 	ID     string // The AppRole  ID
 	Secret string // The Approle secret ID
 	Retry  time.Duration
@@ -41,6 +43,13 @@ type AppRole struct {
 type Store struct {
 	// Addr is the HTTP address of the Vault server.
 	Addr string
+
+	// Engine is the path of the K/V engine to use.
+	//
+	// Vault allows multiple engines of the same type
+	// mounted at the same time and/or engines mounted
+	// at arbitrary paths.
+	Engine string
 
 	// Location is the location on Vault's K/V store
 	// where this KeyStore will save secret keys.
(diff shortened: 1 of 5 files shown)


internal/vault/vault.go
```diff
@@ -196,7 +205,7 @@ func (s *Store) Create(key, value string) error {
 
 	// We try to check whether key exists on the K/V store.
 	// If so, we must not overwrite it.
-	location := fmt.Sprintf("/kv/%s/%s", s.Location, key)
+	location := path.Join(s.Engine, s.Location, key) // /<engine>/<location>/<key>
```
harshavardhana, line 208: Isn't `/` in beginning needed this is not present with path.Join()

aead: The Vault client takes care of this. It will create a request similar to:
```
https://127.0.0.1:8200/secrets/prefix/my-key
```
So,if there is a missing `/` between the endpoint and the path it will insert one (part of request construction).

aead: Otherwise it would also not work if you e.g. specify `https://127.0.0.1:8200/` instead of `https://127.0.0.1:8200`. (Trailing `/`). Either you would have a missing `/` or 2 `/`s.

aead:
ping @harshavardhana

PR #62, opened 2020-06-16
build: fix main instead of dir in gorelease config

harshavardhana:


minisign.pub
```diff
@@ -0,0 +1,2 @@
+untrusted comment: minisign public key 2FD021B6F59AE5F1
```
aead, line 1: Checking of public key is not necessary. Instead README.md must be updated.

PR #63, opened 2020-06-26
gemalto: add support for Gemalto KeySecure

aead:
This commit adds support for Gemalto KeySecure
as KMS.

KeySecure is a KMS implementation developed by Gemalto
(now Thales) and provides a KMIP as well as a REST API.
This commit adds a REST client that uses the KeySecure
secrets API (`/v1/vault/secrets`) to fetch and store
cryptographic keys.

***

KeySecure has multi-user support (username / password).
However, KeySecure also has the concept of refresh and
authentication tokens for programmatic access.

A refresh token is an opaque string, e.g.
`CEvk5cdHLG7si05LReIeDbXE3PKD082YdUFAnxX75md3jzV0BnyHyAmPPJiA0Kgn`,
that can be generated by a user to give an application programmatic
access. With the refresh token an application can request a short-lived
(5 min by default) authentication token. The authentication token
can then be used to perform operations - like creating a new secret.
Therefore, it has to be sent as part of the `Authorization` header.

With the refresh token the application can renew the authentication
token before it expires. KES supports only refresh tokens, not
username/password, since a refresh token is the recommended approach
to grant programmatic access. A refresh token can be easily revoked.

***

Apart from authentication, KeySecure provides a secret store that is
quite similar to e.g. Hashicorp Vault's K/V engine. Therefore, KES
uses the secrets API to store cryptographic keys.

***

This commit also adds a wrapper for an HTTP client that implements an
automatic retry mechanism. This is needed since KeySecure does not
provide a Go SDK and KES should be robust and tolerate temp. network
errors.

### DIFF
--- a/internal/gemalto/key-secure.go
+++ b/internal/gemalto/key-secure.go
@@ -1,0 +1,354 @@
+// Copyright 2020 - MinIO, Inc. All rights reserved.
+// Use of this source code is governed by the AGPLv3
+// license that can be found in the LICENSE file.
+
+// Package gemalto implements a key store that fetches/stores
+// cryptographic keys on a Gemalto KeySecure instance.
+package gemalto
+
+import (
+	"bytes"
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"net"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/minio/kes"
+	xhttp "github.com/minio/kes/internal/http"
+)
+
+// Credentials represents a Gemalto KeySecure
+// refresh token that can be used to obtain a
+// short-lived authentication token.
+//
+// A token is valid within either the default root
+// domain (empty) or a specifc domain - e.g. my-domain.
+type Credentials struct {
+	Token  string        // The KeySecure refresh token
+	Domain string        // The KeySecure domain - similar to a Vault Namespace
+	Retry  time.Duration // The time to wait before trying to re-authenticate
+}
+
+// KeySecure is a Gemalto KeySecure client that
+// stores / fetches key-value pairs as secrets.
+//
+// It tries to connect to a KeySecure instance
+// at the given endpoint and uses the login
+// credentials to authenticate.
+type KeySecure struct {
+	// Endpoint is the KeySecure instance endpoint.
+	Endpoint string
+
+	// CAPath is a path to the root CA certificate(s)
+	// used to verify the TLS certificate of the KeySecure
+	// instance. If empty, the host's root CA set is used.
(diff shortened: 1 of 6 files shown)


internal/gemalto/key-secure.go
```diff
@@ -0,0 +1,354 @@
...
+		},
+	}
+	if err = s.client.Authenticate(s.Endpoint, s.Login); err != nil {
+		return err
+	}
+	go s.client.RenewAuthToken(context.Background(), s.Endpoint, s.Login)
+	return nil
+}
+
+// Create creates the given key-value pair at Vault if and only
```
Alevsk, line 120:
```suggestion
// Create creates the given key-value pair at Gemalto if and only
```

internal/gemalto/key-secure.go
```diff
@@ -0,0 +1,354 @@
...
+
+// Credentials represents a Gemalto KeySecure
+// refresh token that can be used to obtain a
+// short-lived authentication token.
+//
+// A token is valid within either the default root
+// domain (empty) or a specifc domain - e.g. my-domain.
+type Credentials struct {
+	Token  string        // The KeySecure refresh token
+	Domain string        // The KeySecure domain - similar to a Vault Namespace
```
Alevsk, line 38:
```suggestion
	Domain string        // The KeySecure domain - similar to a Gemalto Namespace
```

internal/gemalto/key-secure.go
```diff
@@ -0,0 +1,354 @@
...
+	body, err := json.Marshal(Request{
+		Type:  "seed", // KeySecure supports blob, password and seed
+		Value: value,
+		Name:  key,
+	})
+	if err != nil {
+		return err
+	}
+
+	url := fmt.Sprintf("%s/api/v1/vault/secrets", s.Endpoint)
```
Alevsk, line 139: should this be gemalto? or Gemalto uses vault under the hood?

aead: No, Gemalto KeySecure and Hashicorp Vault are two different (KMS) vendors.

internal/gemalto/key-secure.go
aead: A Gemalto KeySecure domain is kind of equivalent to a Hashicorp Vault Namespace.
A Vault namespace is an internal isolation layer:
```
     Vault
       │
       ├── NS_1 [ users, policies, keys, ...]
       │
       └── NS_2 [ users, policies, keys ...]
```

aead:
> How do we test this?

@harshavardhana You would need access to a Gemalto KeySecure instance.

harshavardhana:
> > How do we test this?
> 
> @harshavardhana You would need access to a Gemalto KeySecure instance.

Can I get access? @aead

harshavardhana:
Let's take this @aead and I shall test it after its merged.

PR #66, opened 2020-07-13
server: add server startup message

aead:
This commit adds a `-q, --quiet` flag to the
`kes server` command.

If not specified the server now prints a startup
message that contains the server endpoint(s),
root identity, key store location a.s.o.

If the `-q` flag is specified the server does
not print a startup message and behaves as before.

### DIFF
--- a/cmd/kes/server.go
+++ b/cmd/kes/server.go
@@ -6,22 +6,27 @@ package main
 
 import (
 	"context"
+	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/hex"
 	"errors"
 	"flag"
 	"fmt"
 	"io/ioutil"
 	stdlog "log"
+	"net"
 	"net/http"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"sync"
 	"syscall"
 	"time"
 
+	"github.com/fatih/color"
 	"github.com/minio/kes"
 	"github.com/minio/kes/internal/auth"
 	"github.com/minio/kes/internal/aws"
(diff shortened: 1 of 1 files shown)


aead:
Screenshot from 2020-07-13 15-28-55

harshavardhana:
If you have lots of local IPs @aead we will take up a lot of space writing it in one line would be sufficient IMO 

> Endpoint:  http://10.0.0.67:9000  http://172.17.0.1:9000  http://172.18.0.1:9000  http://127.0.0.1:9000          

Can the CLI also be written in one line? @aead - currently I can copy-paste it properly 

What is the use for displaying `Root:` for the client? @aead

aead:
> If you have lots of local IPs @aead we will take up a lot of space writing it in one line would be sufficient IMO

The problem with one line is that if you have lots of IPs or don't have a small terminal window (not full screen) the IPs wrap at the line ending which looks really weird. Instead what I can do is write in one line until it would wrap, and then, start a new properly aligned line. Like:
```
Endpoint: https://<IP>:<port>    https://<IP>:<port>   https://<IP>:<port>
          https://<IP>:<port>    https://<IP>:<port>
```

> Can the CLI also be written in one line? @aead - currently I can't copy-paste it properly

You have to edit it anyway, right?! Are you saying have the CLI like:
```
CLI:     export KES_SERVER=https://<ip>:<port> KES_CLIENT_KEY=<client-key> KES_CLIENT_CERT=<client-cert>
```
Again, here it's slightly easier to copy-paste but I cannot properly document for new users (see the `// e.g.` comment)
and there is a again the line wrapping issue.

> What is the use for displaying Root: for the client?

To display who is root and whether a root identity is present or not. If you e.g. start the server with `--root=_`
the you get:
```
Root:     _     [ disabled ]
```

Root in KES is like MinIO admin - so a user may want to see on startup who / whether there is root.

aead:
@harshavardhana PTAL at the list of endpoints [here](https://github.com/minio/kes/pull/66#issuecomment-657564313). Now, there are two endpoints per line. Let me know whether that is acceptable for you.

PR #67, opened 2020-07-21
cli: rework audit log trace and add error log trace

aead:
This commit does two things:
 - Add error log tracing via the `/v1/log/error/trace` API
 - Rework the audit log tracing CLI UI.

Now, the log event stream to STDOUT  is displayed as a table:
```
┌──────────┬────────────┬──────────────────┬───────────────────────────────────┬────────────┐
│ Time     │ Identity   │ Status           │ API Operation                     │ Resp. Time │
├──────────┼────────────┼──────────────────┼───────────────────────────────────┼────────────┤
│ 15:07:50 │ 3ecfcdf38… │ [200 OK]         │ /v1/log/audit/trace               │ 272µs      │
│ 15:08:06 │ 3ecfcdf38… │ [200 OK]         │ /v1/key/generate/my-minio-key     │ 550µs      │
│ 15:08:18 │ 3ecfcdf38… │ [200 OK]         │ /v1/key/decrypt/my-minio-key      │ 392µs      │
└──────────┴────────────┴──────────────────┴───────────────────────────────────┴────────────┘
```

A user can specify whether the error or the audit log should be traced
via the `--type=[audit|error]` flag. The default is `--type=audit` which
is aligned with the previous behavior.

If a user wants to trace error and audit logs at the same time (e.g.
with `--json` flag) the following bash/zsh command can be used:
```
{ kes log trace --type=error --json & kes log trace --type=audit --json; } | jq
```

The current table CLI output requires that the terminal supports unicode
characters - which is the case for modern terminals, including windows.

### DIFF
--- a/cmd/kes/log.go
+++ b/cmd/kes/log.go
@@ -7,14 +7,15 @@ package main
 import (
 	"flag"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"os/signal"
-	"runtime"
-	"syscall"
+	"strings"
 	"time"
 
 	"github.com/fatih/color"
+	"github.com/minio/kes"
 )
 
 const logCmdUsage = `usage: %s <command>
@@ -52,6 +53,11 @@ Connects to a KES server and traces log events.
 
 usage: %s [flags]
 
+  --type               Specify the log event type. (default: audit)
+                       Valid options are:
+                          --type=audit  
+                          --type=error
+
   --json               Print log events as JSON.
 
   -k, --insecure       Skip X.509 certificate validation during TLS handshake.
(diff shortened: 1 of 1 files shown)


aead:
**Windows**
image

**Linux**
Screenshot from 2020-07-21 17-16-39
Screenshot from 2020-07-21 17-16-56

aead:
> Did you explore using more friendlier terminal UIs like what sidekick does?

@harshavardhana Yes, but all of these approaches are not "cross-platform" (i.e. windows) due to the minimal functionality supported by the standard windows terminal.

harshavardhana:
> @harshavardhana Yes, but all of these approaches are not "cross-platform" (i.e. windows) due to the minimal functionality supported by the standard windows terminal.
> The current UI is consistent on linux, macOS and windows. It's based on the approach used by fzf.

@aead you are assuming quite a few things sidekick UI works fine on windows and macOS as well. 

Writing manually all this work is not necessary when there are termUI's available to do this for you, 
the approach is taken here also doesn't scale up and down properly based on terminal windows.

aead:
> you are assuming quite a few things sidekick UI works fine on windows and macOS as well.

Not in the windows default terminal AFAIK. It only works in powershell.

> Writing manually all this work is not necessary when there are termUI's available to do this for you,
the approach is taken here also doesn't scale up and down properly based on terminal windows

The approach has significantly less code then the implementation using github.com/rivo/tview before.
Also, properly scaling the UI is not possible on windows (non-powershell) since you "re-write" previous lines - only the current line.

aead:
Let me revisit https://github.com/gdamore/tcell

PR #68, opened 2020-07-27
cli: rework audit log trace and add error log trace

aead:
This commit does two things:
 - Add error log tracing via the `/v1/log/error/trace` API
 - Rework the audit log tracing CLI UI.

Now, the log event stream to STDOUT is displayed as a table:
```
┌──────────┬─────────────┬─────────────┬─────────────────────────────────────────┬──────────┐
│   Time   │  Identity   │   Status    │ API Operations                          │ Response │
├──────────┼─────────────┼─────────────┼─────────────────────────────────────────┼──────────┤
│ 13:43:48 │ 3ecfcdf38f… │   200 OK    │ /v1/log/audit/trace                     │  282µs   │
│ 15:08:06 │ 3ecfcdf38f… │   200 OK    │ /v1/key/generate/my-minio-key           │  550µs   │
└──────────┴─────────────┴─────────────┴─────────────────────────────────────────┴──────────┘
```

A user can specify whether the error or the audit log should be traced
via the `--type=[audit|error]` flag. The default is `--type=audit` which
is aligned with the previous behavior.

If a user wants to trace error and audit logs at the same time (e.g.
with `--json` flag) the following bash/zsh command can be used:
```
{ kes log trace --type=error --json & kes log trace --type=audit --json; } | jq
```

The current table CLI output requires that the terminal supports unicode
characters - which is the case for modern terminals, including windows.


aead:
**Audit log trace**
Screenshot from 2020-07-27 15-37-43
Screenshot from 2020-07-27 15-37-58

**Error log trace**
Screenshot from 2020-07-27 15-40-56
Screenshot from 2020-07-27 15-41-09

cmd/kes/log.go
```diff
@@ -65,8 +74,13 @@ func logTrace(args []string) error {
...
 
-	var jsonOutput bool
-	var insecureSkipVerify bool
+	var (
+		typeFlag   string
+		jsonOutput bool
+
+		insecureSkipVerify bool
+	)
+	cli.StringVar(&typeFlag, "type", "audit", "Log event type [ audit | error ]")
```
harshavardhana, line 83: is there a good reason for you to use Go flags @aead  and not `minio/cli` ?

aead: The argument works the other way around.
Currently, the Go standard library works for this project. So, there is no need (yet) to add a 3rd party library for this task.

aead:
ping @Alevsk

aead:
ping @Alevsk

cmd/kes/log.go
```diff
@@ -79,60 +93,200 @@ func logTrace(args []string) error {
...
+}
+
+// traceAuditLogWithUI iterates over the audit log
+// event stream and prints a table-like UI to STDOUT.
+//
+// Each event is displayed as a new row and the UI is
+// automatically adjusted to the terminal window size.
+func traceAuditLogWithUI(stream *kes.AuditStream) error {
+	table := xterm.NewTable("Time", "Identity", "Status", "API Operations", "Response")
+	table.Header()[0].Width = 0.12
```
Alevsk, line 140: could `table.Header()` be `empty` or `nil` at some point? that will cause `npe`

cmd/kes/log.go
```diff
@@ -79,60 +93,200 @@ func logTrace(args []string) error {
...
 }
+
+// traceErrorLogWithUI iterates over the error log
+// event stream and prints a table-like UI to STDOUT.
+//
+// Each event is displayed as a new row and the UI is
+// automatically adjusted to the terminal window size.
+func traceErrorLogWithUI(stream *kes.ErrorStream) error {
+	table := xterm.NewTable("Time", "Error")
+	table.Header()[0].Width = 0.12
```
Alevsk, line 221: same question as above

cmd/kes/log.go
aead: It can an `empty` slice. However, it's kind of the callers responsibility to access only table headers that exists.
Here I define 5 columns and therefore 5 headers exists:
```
table := xterm.NewTable("Time", "Identity", "Status", "API Operations", "Response")
```
See: https://github.com/minio/kes/pull/68/files#diff-9e123ecf3fdb4c1d6991d8cb7b908636R71-R72

cmd/kes/log.go
aead: https://github.com/minio/kes/pull/68#discussion_r468875652

PR #78, opened 2020-08-25
add server metrics API

aead:
This commit adds a new server API that exposes network metrics.
The new API endpoint is: `/v1/metrics`

Now, the server gathers networks metrics, like the internal
request-response latency and the number of requests that succeeded
resp. failed, and exposes them via the `v1/metrics` endpoint.

The metrics are exposed in the Prometheus exposition format:
```
 # HELP kes_http_request_error Number of request that failed due to some error. (HTTP 4xx status code)
 # TYPE kes_http_request_error counter
 kes_http_request_error 1
 # HELP kes_http_request_failure Number of request that failed due to some internal failure. (HTTP 5xx status code)
 # TYPE kes_http_request_failure counter
 kes_http_request_failure 0
 # HELP kes_http_request_success Number of requests that have been served successfully.
 # TYPE kes_http_request_success counter
 kes_http_request_success 5
 # HELP kes_http_response_time Histogram of request response times spawning from 10ms to 10s.
 # TYPE kes_http_response_time histogram
 kes_http_response_time_bucket{le="0.01"} 6
 kes_http_response_time_bucket{le="0.05"} 6
 kes_http_response_time_bucket{le="0.1"} 6
 kes_http_response_time_bucket{le="0.25"} 6
 kes_http_response_time_bucket{le="0.5"} 6
 kes_http_response_time_bucket{le="1"} 6
 kes_http_response_time_bucket{le="1.5"} 6
 kes_http_response_time_bucket{le="3"} 6
 kes_http_response_time_bucket{le="5"} 6
 kes_http_response_time_bucket{le="10"} 6
 kes_http_response_time_bucket{le="+Inf"} 6
 kes_http_response_time_sum 0.000371124
 kes_http_response_time_count 6
```

Now, the server can be monitored by a Prometheus instance that
authenticates via mTLS (like any other client) and scrapes
the `/v1/metrics` API.

The metrics API can be accessed e.g. via cURL:
```
curl -sSL -X GET --key <key> --cert <cert> https://127.0.0.1:7373/v1/metrics -k
```


aead:
ping @harshavardhana @dvaldivia

internal/metric/metric.go
```diff
@@ -0,0 +1,196 @@
...
+var (
+	_ http.ResponseWriter = (*countResponseWriter)(nil)
+	_ http.Flusher        = (*countResponseWriter)(nil)
+)
+
+func (w *countResponseWriter) WriteHeader(status int) {
+	w.ResponseWriter.WriteHeader(status)
+	if !w.written {
+		switch {
+		case status == http.StatusOK:
```
Alevsk, line 173:
```suggestion
		case status == http.StatusOK || status == http.StatusNoContent:
```
should we check for `ok no content` status as well?, there is lot more of ok responses too:

```
StatusOK                   = 200 // RFC 7231, 6.3.1
	StatusCreated              = 201 // RFC 7231, 6.3.2
	StatusAccepted             = 202 // RFC 7231, 6.3.3
	StatusNonAuthoritativeInfo = 203 // RFC 7231, 6.3.4
	StatusNoContent            = 204 // RFC 7231, 6.3.5
	StatusResetContent         = 205 // RFC 7231, 6.3.6
	StatusPartialContent       = 206 // RFC 7233, 4.1
	StatusMultiStatus          = 207 // RFC 4918, 11.1
	StatusAlreadyReported      = 208 // RFC 5842, 7.1
	StatusIMUsed               = 226 // RFC 3229, 10.4.1
```

aead: That is correct. However, no `http.HandlerFunc` implemented by KES returns anything else (or is supposed to return anything else)  then `200 OK`. If that should ever change it should be tracked as well - maybe separately. To me e.g. `PartialContent` or `IMUsed` is quite different from `OK`.

If someone writes a`http.HandlerFunc` that returns something other then `200 OK` in case of success but forgets to update the metrics then we fail early:
```
			// We panic to signal that the server returned a status code
			// that is not tracked. If, in the future, the application
			// returns a new (kind of) status code it should be collected
			// as well.
			// Otherwise, we would silently ignore new status codes and the
			// metrics would be incomplete.
			panic("metrics: unexpected response status code " + strconv.Itoa(status))
```
(see a few lines below)

Alevsk (dismissed):

Alevsk (approved):

aead:
ping @harshavardhana

harshavardhana (approved):

PR #79, opened 2020-09-02
feat: add GCP secret manager key store

ymchun:
This PR included the following changes

- add gcp key store config to `server-config.yaml`
- add `internal/gcp/secret-manager.go`
- fix some `server-config.yaml` indent problem
- fix some mis-spelling words
- remove tailing whitespace


cmd/kes/server.go
```diff
@@ -129,19 +130,25 @@ func server(args []string) error {
...
+		keyStoreConfigCount++
+	}
+	if config.Keys.Gemalto.KeySecure.Endpoint != "" {
+		keyStoreConfigCount++
+	}
+	if config.Keys.GCP.ProjectID != "" {
+		keyStoreConfigCount++
+	}
+	if keyStoreConfigCount > 1 {
+		return errors.New("Ambiguous configuration: multiple key store are specified at the same time")
```
harshavardhana, line 151: Looks like you have removed the context from the error, now we have to guess what the issue is about what was wrong with the previous per target errors?

internal/http/retry.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 // RetryReader returns an io.ReadSeeker that can be
-// used as request body for retriable requests via
+// used as request body for retirable requests via
```
harshavardhana, line 20:
```suggestion
// used as request body for retryable requests via
```

internal/secret/secret.go
```diff
@@ -128,7 +128,7 @@ func (s Secret) Wrap(plaintext, associatedData []byte) ([]byte, error) {
 // returns the resuting plaintext. It returns an
 // error if ciphertext is malformed or not authentic.
 func (s Secret) Unwrap(ciphertext []byte, associatedData []byte) ([]byte, error) {
-	// TODO(aead): The Go JSON unmarshaling is malleable.
+	// TODO(aead): The Go JSON un-marshaling is malleable.
 	// For instance, it ignores the first key-value pair if
 	// the same key is present more than nonce or ignores
```
harshavardhana, line 133:
```suggestion
	// the same key is present more than once or ignores
```

cmd/kes/server.go
ymchun: Yes, you're right. Let me revert this part.

internal/gcp/secret-manager.go
```diff
@@ -0,0 +1,176 @@
+// Copyright 2019 - MinIO, Inc. All rights reserved.
```
harshavardhana, line 1:
```suggestion
// Copyright 2020 - MinIO, Inc. All rights reserved.
```

harshavardhana (changes requested):
Rest of the changes LGTM

cmd/kes/config.go
```diff
@@ -104,6 +104,11 @@ type serverConfig struct {
 				} `yaml:"tls"`
 			} `yaml:"keysecure"`
 		} `yaml:"gemalto"`
+
+		GCP struct {
```
aead, line 108: This configuration should follow the AWS secrets manager configuration:
```
GCP struct {
   SecretsManager struct {
       Endpoint string `yaml:"endpoint"`
       ...
   }
}
```

We want to also be able to support any GCP secretsmanager compatible implementation.

cmd/kes/config.go
```diff
@@ -104,6 +104,11 @@ type serverConfig struct {
...
+			ProjectID       string `yaml:"projectid"`
+			CredentialsFile string `yaml:"credentialsfile"`
```
aead, line 110: I'm not in favor of requiring another file for the credentials.
The config file then is no longer self-contained. Instead, the credentials should be specified here. Eventually, we have to
"copy" the GCP auth. key structure here:
```
{
  "name": "projects/project-id/serviceAccounts/service-account-email/keys/key-id",
  "privateKeyData": "private-key",
  "validAfterTime": "date",
  "validBeforeTime": "date",
  "keyAlgorithm": "KEY_ALG_RSA_2048"
}
(See: https://cloud.google.com/iam/docs/creating-managing-service-account-keys)
I have to explore more about the general GCP auth. mechanisms to make a informed decision...
```

internal/gcp/secret-manager.go
```diff
@@ -0,0 +1,176 @@
...
+	"net/http"
+
+	secretmanager "cloud.google.com/go/secretmanager/apiv1"
+	"github.com/minio/kes"
+	"google.golang.org/api/option"
+	secretmanagerpb "google.golang.org/genproto/googleapis/cloud/secretmanager/v1"
+)
+
+type SecretManager struct {
+	// The GCP project id
```
aead, line 20:
```suggestion
	// The project ID is a unique, user-assigned ID that can be used by Google APIs.
	// The project ID must be a unique string of 6 to 30 lowercase letters, digits, or hyphens.
	// It must start with a letter, and cannot have a trailing hyphen.
	// See: https://cloud.google.com/resource-manager/docs/creating-managing-projects
```

This has to be more elaborate

internal/gcp/secret-manager.go
```diff
@@ -0,0 +1,176 @@
...
+// GCP secret manager first need to create the secret itself
+// and then add a secret version which is where the actual key-value pair stored
+// The secret will be stored in encrypted form using AES-256
+func (s *SecretManager) Create(key, value string) error {
+	if s.client == nil {
+		s.logf("gcp: no connection to GCP secret manager: '%s'", s.ProjectID)
+		return errCreateKey
+	}
+
+	existingSecret, _ := s.client.GetSecret(
```
aead, line 58: We must not ignore the error here. There are to cases to consider. Eventually, GCP returns a well-defined error if a secret is already present in which case we should return `kes.ErrKeyExists`

Also, if another error occurs we have to log that error to the `ErrorLog` / `STDERR`. Otherwise, a user has no way to determine why it's KES-GCP setup does not work. Also, we should return an error with 502 Bad Gateway status code in this case. See e.g. https://github.com/minio/kes/blob/a918c440d5b631a0d2628ae3aec289b6d4cefee1/internal/aws/secrets-manager.go#L91

internal/gcp/secret-manager.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// Get returns the value associated with the given key.
+func (s *SecretManager) Get(key string) (string, error) {
+	if s.client == nil {
+		s.logf("gcp: no connection to GCP secret manager: '%s'", s.ProjectID)
+		return "", errGetKey
+	}
+
+	result, err := s.client.AccessSecretVersion(
```
aead, line 114: What happens if the secret does not exist. Is there no way to determine whether this is the case and return `kes.ErrKeyNotFound`

internal/gcp/secret-manager.go
```diff
@@ -0,0 +1,176 @@
...
+func (s *SecretManager) logf(format string, v ...interface{}) {
+	if s.ErrorLog == nil {
+		log.Printf(format, v...)
+	} else {
+		s.ErrorLog.Printf(format, v...)
+	}
+}
+
+func (s *SecretManager) getSecretName(projectID, secretID string) string {
+	return fmt.Sprintf("projects/%s/secrets/%s/versions/1", projectID, secretID)
```
aead, line 175:
```suggestion
	return path.Join("projects", projectID, "secrets", secretID, "versions/1")
```
Handles path segments properly...

aead (changes requested):
@ymchun Thank you very much for this PR.

ymchun:
@aead Thanks for your review! I've updated the code according to your change requests.
For the credentials file, we may need some time to figure out a better way to handle it.
In my point of view, the JSON credentials file is directly created and downloaded from gcp console and we shouldn't change it.

aead:
> For the credentials file, we may need some time to figure out a better way to handle it.
In my point of view, the JSON credentials file is directly created and downloaded from gcp console and we shouldn't change it.

@ymchun Yes, the credentials file is provided by GCP when creating e.g. a service account.
I'm not really suggesting to change it. I'm currently testing this integration and see what exactly GCP expects from the client.
My idea is that the values (a subset of required fields) of JSON credentials file are specified as part of the the KES GCP configuration.
Such that the KES config file is self-contained and does not depend on another file. I'm currently investigating this and will give an update. Give one more day :smile:

ymchun:
@aead Thank you very much! I agreed that the config file should be self-contained. Copying content from JSON file looks good to me as well. My concern is that we may need to bear the risk when Google change the JSON file structure but as long as we keep using the same version of client library, it should have no problem.

aead:
@ymchun I've reworked the implementation a bit to make the config file self-contained and changed a few minor details.
Can you please update your PR with the following patch - e.g. via:
```
git apply --stat secretmanager.txt
git apply --check secretmanager.txt
git am --signoff < secretmanager.txt
```

[secretmanager.txt](https://github.com/minio/kes/files/5233224/secretmanager.txt)

ymchun:
@aead thank you! updated the PR

error.go
```diff
@@ -78,7 +78,7 @@ func (e Error) Error() string { return e.message }
 //
 // If the response status code is < 400, e.g. 200 OK,
 // parseErrorResponse returns nil and does not attempt
-// to read or close the responde body.
+// to read or close the responded body.
```
aead, line 81:
```suggestion
// to read or close the response body.
```

aead (commented):
One minor spelling remark. Apart from that LGTM. Thank you very much @ymchun! 
Can you squash the commits for addressing the review comments to just the two commits:
 - 2f3e165 (fix misspelling words and remove tailing space)
 - dd87af5 (feat: add gcp secret manager support)
Such that the commit history is clean and the GCP feature can be linked to one commits.

ymchun:
All done! thank you for your review. @aead @harshavardhana

aead (approved):
LGTM :heart:

harshavardhana (approved):

PR #81, opened 2020-10-22
adjust CLI to be more like modern unix commands

aead:
This commit refactors the CLI help messages
and error reporting to follow modern unix command
styles.

In particular, the command help messages contains
a list of options, an short description and a example
showing the command.

Further, this commit unifies the error messages to start
with: `Error:` to visually distinguish error messages from
command results.


harshavardhana (commented):
I see that you have implemented everything from barebones, why not use `minio/cli` ? @aead is this because of licensing constraints?

aead:
> I see that you have implemented everything from barebones, why not use minio/cli ? @aead is this because of licensing constraints?

I think there is no license issue. However, we're trying to avoid using external libraries when similar /equivalent functionality is available within the standard library and when using external libraries would not lead to a "significant" reduction of complexity.
This is an approach followed by popular (CLI) projects - like [fzf](https://github.com/junegunn/fzf/blob/master/src/options.go#L19) or [age](https://github.com/FiloSottile/age/blob/master/cmd/age/age.go#L33). It tries to keep entry-barrier for new contributors low.

Similarly, the KES server implementation does not use a "web framework" - like [gorilla/mux](https://github.com/gorilla/mux) - for its REST API but plain `net/http`.

Alevsk (approved):

harshavardhana (approved):

PR #82, opened 2020-11-12
add new server API for listing keys

aead:
This commit adds a new server API:
```
/v1/key/list/<pattern>
```

This API returns a nd-json stream of
`KeyDescription` objects - each one
describing one key at the KES server.

The key listing API is designed to have
a low memory footprint (since there may
be thousands of keys) even when many
clients try to list keys in parallel.

Therefore, the listing implementation
(at the KES server as well as at the client)
uses iterators to list keys lazy. As soon
as a client disconnects / times out /...
the listing is aborted to not waste server
resources.

This commit also adds a CLI command:
```
kes key list [options] [<pattern>]
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -6,6 +6,7 @@ package kes
 
 import (
 	"bytes"
+	"context"
 	"crypto/tls"
 	"encoding"
 	"encoding/base64"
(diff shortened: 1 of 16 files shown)


aead:
Do-Not-Merge: Waiting for a new release using Go `1.15.5` addressing: https://groups.google.com/g/golang-announce/c/NpBGTTmKzpM/m/fLguyiM2CAAJ

Alevsk (approved):

PR #85, opened 2020-11-25
cli: add `kes tool migrate` command

aead:
This commit adds a new CLI command:
```
kes tool migrate

Usage:
    kes tool migrate [options] [<pattern>]

Options:
    --from <PATH>          Path to the configuration file of the server that
                           should be migrated
    --to   <PATH>          Path to the configuration file of the server that
                           is the migration target

    -f, --force            Migrate keys even if a key with the same name exists
                           at the target. The existing keys will be deleted

    --merge                Merge the source into the target by only migrating
                           those keys that do not exist at the target

    -q, --quiet            Don't print migration progress and statistics.
    -h, --help             Show list of command-line options

Examples:
    $ kes tool migrate --from kes-vault.yml --to kes-aws.yml
```

With the `kes tool migrate` command it is possible to move some/all
keys from one KMS backend (e.g. Hashicorp Vault) to another KMS backend
(e.g. AWS SecretsManager).

However, the command does not use a KES server, and therefore,
does not require a running source nor destination KES server instance.

Instead, `kes tool migrate` consumes two KES config files (for the
source and the target) and then migrates the keys by directly talking
to the KMS backends specified in the config files.

This approach has the following implications:
 - Since `kes tool migrate` operates directly on the KMS backends
   it does not require a running KES instance as source nor
   as target.
 - The KES server does not have to provide an API to extract keys.
   Such an API would be an interesting target for attacks since it
   would allow an successful attacker to copy all keys from the
   KMS. Therefore, we tried to avoid implementing such an API.
 - The `kes tool migrate` command has to be run on a machine that
   can reach the source KMS as well as the target KMS.

### DIFF
--- a/cmd/kes/config.go
+++ b/cmd/kes/config.go
@@ -5,11 +5,23 @@
 package main
 
 import (
+	"context"
+	"errors"
+	"fmt"
+	stdlog "log"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
 	"github.com/minio/kes"
+	"github.com/minio/kes/internal/aws"
+	"github.com/minio/kes/internal/fs"
+	"github.com/minio/kes/internal/gcp"
+	"github.com/minio/kes/internal/gemalto"
+	"github.com/minio/kes/internal/mem"
+	"github.com/minio/kes/internal/secret"
+	"github.com/minio/kes/internal/vault"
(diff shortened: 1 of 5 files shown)


Alevsk (approved):
LGTM

PR #88, opened 2020-11-27
Add Dockerfile with Red Hat UBI

nitisht:


Dockerfile.release.ubi
```diff
@@ -0,0 +1,22 @@
...
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.3
+
+COPY CREDITS /licenses/CREDITS
+COPY LICENSE /licenses/LICENSE
+
+LABEL name="MinIO" \
+      vendor="MinIO Inc <dev@min.io>" \
+      maintainer="MinIO Inc <dev@min.io>" \
+      version="v0.12.1" \
```
aead, line 12: Should the label include the version and release information?!
Since that would require us to update the docker file with a commit before making a release every time...

harshavardhana: Yes we need to @aead

harshavardhana: Also every release we don't need to make a UBI release

aead: Okay

aead (approved):

harshavardhana (approved):

PR #89, opened 2020-11-30
refactor the logging infrastructure

aead:
This commit refactors the server logging implementation.

The `AuditResponseWriter` and `FlushWriter` are moved
from `internal/log` to `internal/http`.

The `AuditResponseWriter` is simplified. Now, it uses
the `kes.AuditEvent` type instead of creating a compatible
JSON encoding itself. Further, it now an audit event contains
the point in time when the server received the request (as documented)
not when it produced the response.

The `JSONWriter` becomes the `ErrEncoder`. Its implementation now
uses `json.Encoder` instead of crafting nd-json by hand.

The `log.SystemLog` becomes `log.Target` and its implementation is
more robust now.


Alevsk (approved):
LGTM

PR #90, opened 2020-12-02
vault: add support for kubernetes authentication method

aead:
This commit adds support for the Vault kubernetes authentication
method: https://www.vaultproject.io/docs/auth/kubernetes

Now, the KES configuration can contain the following K8S auth section:
```
kubernetes:
  engine: ""
  role:   ""
  jwt:    ""
  retry:  15s
```

The `engine` allows the customization of the auth path configured
at Vault. It defaults to the Vault default: `kubernetes`.

The `retry` is the delay KES waits before trying to re-authenticate
after connection loss.

The `role` and `jwt` are the credentials received from K8S.

Currently, KES does not support both auth methods (`approle` | `kubernetes`)
at the same time. So, a operator must not set AppRole and Kubernetes
credentials.

### DIFF
--- a/cmd/kes/config.go
+++ b/cmd/kes/config.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	stdlog "log"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


aead:
do-not-merge: The service account JWT is typically mounted as separate secret. This requires some adjustments to how the k8s credentials are passed.

aead:
> if its a path just read from it, if not use the string as JWT @aead

That's the plain :wink:  Just need to figure out whether the role can be part of the token file or not.

Alevsk (approved):

harshavardhana (approved):

PR #91, opened 2020-12-04
config: allow most config fields be replaced with env. vars

aead:
This commit adds support for replacing most config fields
with environment variables.

Now, it is possible to use e.g. the following config snippet:
```
keys:
  vault:
    endpoint: ${VAULT_ENDPOINT}
    approle:
      id: ${VAULT_APPROLE_ID}
      secret: ${VAULT_APPROLE_SECRET}
...
```

Before, only identities could be overwritten with env. variables.

However, there are two kind of fields that cannot be overwritten
with env. variables:
 - Durations, like expiry.
 - Policy paths, like `/v1/key/create/my-key`


Alevsk (dismissed):
LGTM

Alevsk (approved):

PR #92, opened 2020-12-07
Remove alpine base image and use UBI as default

nitisht:


Dockerfile.release
```diff
@@ -1,12 +1,24 @@
 FROM alpine:latest as alpine
 RUN apk add -U --no-cache ca-certificates
```
harshavardhana, line 3: We don't need this @nitisht - we can use only one layer here i.e RedHat UBI image

harshavardhana (changes requested):

harshavardhana (approved):

aead (approved):

PR #95, opened 2021-01-08
fix incorrect lookup of Vault TLS config

aead:
This commit fixes a bug when merging the
KES configuration with env. variables.

Before, the Vault TLS client certificate
and CA path got replaced by the TLS private
key and TLS client certificate.

Now, the Vault TLS client certificate
and CA path get replaced by the value of
the correct env. variable.


aead:
When this PR is merged we will make a patch release: `v0.13.1` @harshavardhana

harshavardhana (approved):

PR #96, opened 2021-01-11
feat: implement kes cli to selfupdate for new releases

harshavardhana:
This PR adds a way for kes CLI to self-update from GitHub releases directly.
Also updates the binary signature automatically, in future we can add 
third party mirrors as well. 

```sh
~ ./kes -u
Downloading: [=================================================================================================================================================================] 12.62 MiB p/s
Updated ./kes to latest release v0.13.1
```


cmd/kes/update.go
```diff
@@ -0,0 +1,120 @@
...
+		DialContext: (&net.Dialer{
+			Timeout:   timeout,
+			KeepAlive: timeout,
+			DualStack: true,
+		}).DialContext,
+		IdleConnTimeout:       timeout,
+		TLSHandshakeTimeout:   timeout,
+		ExpectContinueTimeout: timeout,
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true,
```
aead, line 35: We must not disable X.509 verification

cmd/kes/update.go
```diff
@@ -0,0 +1,120 @@
...
+	"net/http"
+	"os"
+	"runtime"
+	"time"
+
+	"github.com/cheggaaa/pb/v3"
+	"github.com/minio/selfupdate"
+)
+
+func getUpdateTransport(timeout time.Duration) http.RoundTripper {
```
aead, line 23: Why are we using the same timeout for everything? That does not seem reasonable

cmd/kes/update.go
```diff
@@ -0,0 +1,120 @@
...
+	if err = selfupdate.Apply(barReader, opts); err != nil {
+		bar.Finish()
+		if rerr := selfupdate.RollbackError(err); rerr != nil {
+			return rerr
+		}
+		return err
+	}
+
+	bar.Finish()
+	fmt.Printf("Updated %s to latest release %s\n", os.Args[0], rel)
```
aead, line 118: Maybe provide the binary name as parameter instead of using `os.Args`

cmd/kes/main.go
```diff
@@ -43,14 +44,25 @@ func main() {
...
 	cli.Parse(os.Args[1:])
 
 	if showVersion {
 		fmt.Println("kes version", version)
 		return
 	}
+
+	if selfUpdate {
+		if err := update(); err != nil {
+			stdlog.Fatalln(err)
```
aead, line 61: In case of an error, KES clearly reports that an error occurred:
`stdlog.Fataln("Error: %v", err)`

aead (changes requested):

cmd/kes/main.go
```diff
@@ -32,6 +32,7 @@ Commands:
     tool                     Run specific key and identity management tools.
 
     -v, --version            Print version information
+    -u, --selfupdate         Self updates kes
```
aead, line 35:
```suggestion
    -u, --update         Update to the latest stable release
```

cmd/kes/update.go
harshavardhana: There is no reasonable reason to have different timeouts for each since its short-lived transport. All we need is a top level timeout.

aead (approved):

PR #97, opened 2021-01-12
add metrics support to client SDK

aead:
This commit adds support for KES server
metrics to the client SDK.

Now, a client can fetch the KES server
metrics via the `client.Metrics()` method.

Further, this commit prevents the KES server
metrics from being updated when the metrics
are scrapped. This avoids misleading statistics
that include the metrics gathering as regular
traffic.

### DIFF
--- a/client.go
+++ b/client.go
@@ -11,6 +11,7 @@ import (
 	"encoding"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"io"
 	"net"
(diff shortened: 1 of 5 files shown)


harshavardhana (approved):
Actually this is something we should add `pkg/madmin` to have the support as well 

@aead @kerneltime - also we can fix the metrics prefixes to be `minio_` for all metrics
in the newer change.

PR #98, opened 2021-01-14
do not update once updated to latest release

harshavardhana:
avoid repeatedly updating `kes` if already on
the latest version
    
```
kes -u
```

### DIFF
--- a/cmd/kes/update.go
+++ b/cmd/kes/update.go
@@ -80,6 +80,11 @@ func updateInplace() error {
 		return err
 	}
 
+	if rel == version {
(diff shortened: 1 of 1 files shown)


cmd/kes/update.go
```diff
@@ -80,6 +80,11 @@ func updateInplace() error {
 		return err
 	}
 
+	if rel == version {
+		fmt.Printf("You are already running the most recent version of 'kes'.\n")
```
aead, line 84:
```suggestion
		fmt.Printf("You are already running the latest version %q.\n", version)
```

Eventually, tell me what the latest version is 😃

aead (approved):

Issue #100, opened 2021-01-15
docker: KES binary does not report expected version number


PhilippeAccorsi:
#### Bug describtion

We test Minio with kes server and that working good. Now we tried to use API for monitoring (and seen what can be done with API) and we seen a wrong version number when we execute this command curl --key metric.key --cert metric.cert --request GET 'https://kes:7373/version' -k from client (kes is set in /etc/hosts to use my VM where KES serveur running):
~~~
{"version":"0.0.0-dev"}
~~~

- We use docker images `minio/kes:v0.13.0`.

#### Expected behavior

API request for `/version` show the real version running

#### Additional context

We use Minio and KES server with docker images.

aead:
@PhilippeAccorsi What version does the docker container print when you run e.g. `docker run <image-id> --version`
Does it show `v0.13.0` or `0.0.0-dev`?

PhilippeAccorsi:
Thanks for your answer and your investiguation :)

aead:
#101

aead:
Closing this as incorrect fix as discussed with @harshavardhana

aead:
@PhilippeAccorsi Can you try the patch release `v0.13.1` - i.e:
```
docker run minio/kes:v0.13.1 --version
```
It seems like this issue is already resolved.

PhilippeAccorsi:
> @PhilippeAccorsi Can you try the patch release `v0.13.1` - i.e:
> 
> ```
> docker run minio/kes:v0.13.1 --version
> ```
> 
> It seems like this issue is already resolved.

Yes this is working good :+1: 

Thanks for your answer :)

PhilippeAccorsi:
We just have tested the v0.13.0, but I think we need to test the last version before deploying this on production.

PhilippeAccorsi:
I close this issue because fix is already done since v0.13.1

PR #101, opened 2021-01-18
docker: fix versioning in container images

aead:
This commit address a problem w.r.t. versioning
for docker containers. Now, the container fetches the
binary with the correct version and for the right
architecture on startup.

Before, the container reported an unexpected version
when executing `docker run <id> --version`.

The main problem was that binary versions were not aligned
with docker image tags. A container running the image
`v0.13.0` should also run the binary `v0.13.0` - not the
latest binary version.

### DIFF
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -3,10 +3,12 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal:8.3
 LABEL name="MinIO" \
       vendor="MinIO Inc <dev@min.io>" \
       maintainer="MinIO Inc <dev@min.io>" \
-      version="v0.12.1" \
-      release="v0.12.1" \
+      version="v0.13.3" \
+      release="v0.13.3" \
       summary="KES is a stateless and distributed key-management system for high-performance applications." \
       description="KES as the bridge between modern applications - running as containers on Kubernetes - and centralized KMS solutions. Therefore, KES has been designed to be simple, scalable and secure by default. It has just a few knobs to tweak instead of a complex configuration and does not require a deep understanding of secure key-management or cryptography."
+
+ENV VERSION=v0.13.3
 
 RUN \
     microdnf update --nodocs && \
@@ -14,9 +16,9 @@ RUN \
     microdnf clean all && \
     mkdir /licenses && \
     curl -s -q https://raw.githubusercontent.com/minio/kes/master/CREDITS -o /licenses/CREDITS && \
-    curl -s -q https://raw.githubusercontent.com/minio/kes/master/LICENSE -o /licenses/LICENSE
-
-COPY kes /kes
+    curl -s -q https://raw.githubusercontent.com/minio/kes/master/LICENSE -o /licenses/LICENSE && \
+    curl -sSL -q https://github.com/minio/kes/releases/download/$VERSION/kes-linux-amd64 -o kes && \
+    chmod +x kes
 
 EXPOSE 7373
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.14-alpine as build
+FROM golang:1.15-alpine as build
 
 LABEL maintainer="MinIO Inc <dev@min.io>"
 


harshavardhana (approved):

nitisht (approved):
LGTM & Tested.

Dockerfile.release
```diff
@@ -3,20 +3,22 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal:8.3
...
     microdnf update --nodocs && \
     microdnf install ca-certificates --nodocs && \
     microdnf clean all && \
     mkdir /licenses && \
     curl -s -q https://raw.githubusercontent.com/minio/kes/master/CREDITS -o /licenses/CREDITS && \
-    curl -s -q https://raw.githubusercontent.com/minio/kes/master/LICENSE -o /licenses/LICENSE
-
-COPY kes /kes
+    curl -s -q https://raw.githubusercontent.com/minio/kes/master/LICENSE -o /licenses/LICENSE && \
+    curl -sSL -q https://github.com/minio/kes/releases/download/$VERSION/kes-linux-amd64 -o kes && \
```
harshavardhana, line 20: Well the build is `goreleaser` based so binary is not yet uploaded to Github before container is built @aead so `COPY kes /kes` is correct

harshavardhana (changes requested):

Dockerfile.release
nitisht: But the binary seemed to be from master inside the released container image, ref #100 . Some issue in goreleaser?

aead: I would prefer a image that contains the binary. My (ideal) expectation would be:
 - The image is self-contained and just contains the absolute minimal dependencies to start the binary. Ideally a scratch container but it seems like RH doesn't provide such a UBI base.
 - The KES binary version matches the image tag. So, when I start a container with the image `v0.13.0` the KES binary should be the release `v0.13.0` - not latest or anything else.

Currently, the binary that is included is is not the release that we create. 

***

A similar problem exists in the MinIO docker file. There we download the latest release - regardless what the container image tag is. See: https://github.com/minio/minio/blob/master/Dockerfile.release#L31
That has two problems:
 - The UBI LABEL https://github.com/minio/minio/blob/master/Dockerfile.release#L8 may not reflect what is actually deployed.
 - The binary that is run, has nothing to do with the release tagged by the image.

harshavardhana: The binary uploaded is the binary built locally @aead - we will fix all of that properly the container is not fixed properly in this PR since you don't have access to the release mechanisms.

harshavardhana: > A similar problem exists in the MinIO docker file. There we download the latest release - regardless what the container image tag is. See: https://github.com/minio/minio/blob/master/Dockerfile.release#L31

MinIO's problem is different release style is different not the same problem.

harshavardhana: No one really should be building their own containers that are released - even in case of MinIO it is always built during release.

harshavardhana:
This PR is also solving a non-problem - this is already handled in `goreleaser` - the image not getting updated during the release process was a problem that will be fixed from next release onwards.

```
~ docker run --rm minio/kes:latest --version
kes version 0.13.3
```

I don't think this PR would be needed.

harshavardhana (changes requested):
Only change needed here is

```diff
git diff
diff --git a/Dockerfile.release b/Dockerfile.release
index c6bf137..357c3d6 100644
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -3,8 +3,8 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal:8.3
 LABEL name="MinIO" \
       vendor="MinIO Inc <dev@min.io>" \
       maintainer="MinIO Inc <dev@min.io>" \
-      version="v0.12.1" \
-      release="v0.12.1" \
+      version="v0.13.3" \
+      release="v0.13.3" \
       summary="KES is a stateless and distributed key-management system for high-performance applications." \
       description="KES as the bridge between modern applications - running as containers on Kubernetes - and centralized KMS solutions. Therefore, KES has been designed to be simple, scalable and secure by default. It has just a few knobs to tweak instead of a complex configuration and does not require a deep understanding of secure key-management or cryptography."
 
```

aead:
Closing this as an incorrect fix as discussed with @harshavardhana

PR #104, opened 2021-01-21
add cancellation support to KES SDK and CLI

aead:
This commit adds cancellation support to the
KES SDK and CLI by adding a `context.Context`
parameter to the SDK API functions.

Now, cancelling the context terminates any
request to the KES server early.

Also, the CLI now cancels any operation when
the OS sends an interrupt or kill signal.


Alevsk (approved):
LGTM

harshavardhana (approved):

Issue #105, opened 2021-01-25
Unable to operate KES server inside a docker container


Kinchkun:
#### Bug describtion

When I start kes as docker container with the following command:

```
docker run --rm -p 7373:7373 \
 --name kes-server \
  -v $PWD/certs/:/certs minio/kes:latest \ 
   server --key /certs/server.key --cert /certs/server.crt --root ${ROOT_ID} --auth=off 
```

a tcp connection between client and server is not possible because the server side closes the connection:

```
export KES_SERVER=https://localhost:7373
export KES_CLIENT_CERT=/home/noroot/kes/root.crt
export KES_CLIENT_KEY=/home/noroot/kes/root.key

$ kes key list
Error: failed to list keys matching "*": Get "https://localhost:7373/v1/key/list/%2A": read tcp 127.0.0.1:57106->127.0.0.1:7373: read: connection reset by peer
```

However, the very same command and keys works as expected outside a docker container

server:
```
kes  server --key /home/noroot/kes/certs/server.key --cert /home/noroot/kes//certs/server.crt --root ${ROOT_ID} --auth=off 
```

client command:
```
$ kes key list
```

#### Expected behavior

The docker container opens port 7373 and accepts tcp connections.

#### Additional context

1. What version of Go are you using (`go version`)?

I didn't compile the binaries. I use the dockerhub kes image and the current release 0.13.4 binaries.

2. What operating system and processor architecture are you using (`go env`)?

I reproduced the behavior unter linux and macos. Both use x86_64.

Linux:
```
Linux APFEL 5.8.0-36-generic #40~20.04.1-Ubuntu SMP Wed Jan 6 10:15:55 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```
Macos:

```
Darwin Macbook6.local 20.2.0 Darwin Kernel Version 20.2.0: Wed Dec  2 20:39:59 PST 2020; root:xnu-7195.60.75~1/RELEASE_X86_64 x86_64
```

Thanks

aead:
@Kinchkun The problem is most probably that you don't specify the address the server should be bound to. Therefore, KES defaults to `127.0.0.1:7373`. Now, you host does not pick up the binding (since port 7373 on 127.0.0.1 is already bound).

The following command should work:
```
docker run --rm -p 7373:7373 \
 --name kes-server \
  -v $PWD/certs/:/certs minio/kes:latest \ 
[...]

PR #106, opened 2021-01-26
server: listen on all net interfaces by default

aead:
This commit changes the default behavior of the
server address binding. Before, if no TCP address
(IP:PORT) was specified, the server used `127.0.0.1:7373`.
So, the server would try to listen on port `7373` on
the link-local network interface.

While this behavior is not "wrong" it can lead to confusion
when there are multiple network interfaces and another service
is already listening on the port `7373` - e.g. in case of
docker in combination with port-forwarding:
(host:7373 -> container:7373).

In such a case, the KES server would not be able to listen on
port `7373` on the link-local interface - since it'S already used.
Therefore, an application trying to access the server would experience
errors as if the server is not available:
```
Get "https://localhost:7373/version: read tcp 127.0.0.1:57106->127.0.0.1:7373: read: connection reset by peer
```

This commit changes the default behavior to listen on all network
interfaces (`0.0.0.0`) instead of the link-local address (`127.0.0.1`)

A user who wants to specifically listen only on `127.0.0.1` can always
do that via `--addr=127.0.0.1:7373` or via the config file.

### DIFF
--- a/cmd/kes/server.go
+++ b/cmd/kes/server.go
@@ -37,7 +37,7 @@ const serverCmdUsage = `Usage:
     kes server [options]
 
 Options:
-    --addr <IP:PORT>       The address of the server (default: 127.0.0.1:7373)
+    --addr <IP:PORT>       The address of the server (default: 0.0.0.0:7373)
     --config <PATH>        Path to the server configuration file
     --root  <IDENTITY>     The identity of root - who can perform any operation.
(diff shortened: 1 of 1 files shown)


aead:
Ref #105

Alevsk (approved):

harshavardhana (approved):

PR #111, opened 2021-03-05
add mechanism to create keys in a declarative way

aead:
This commit adds a mechanism to create keys in
a declarative way. Now, keys can be specified
in the `keys` section of the config file:

```
keys:
  - name: my-key
  - name: my-key2
```

The KES server will create these keys before
startup.

This commit is a breaking change. The `keys`
section was used to define the KMS/KeyStore
backend. Now, the KMS backend can be specified
in the `keystore` section.

### DIFF
--- a/cmd/kes/config.go
+++ b/cmd/kes/config.go
@@ -58,7 +58,11 @@ type serverConfig struct {
 		Audit string `yaml:"audit"`
 	} `yaml:"log"`
 
-	Keys kmsServerConfig `yaml:"keys"`
+	Keys []struct {
+		Name string `yaml:"name"`
+	} `yaml:"keys"`
+
+	KeyStore kmsServerConfig `yaml:"keystore"`
 }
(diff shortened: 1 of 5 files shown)


aead:
This PR requires some changes to the docs

dvaldivia:
is there a migration path from previos config to this new config so that operator can update the tenants safely?

aead:
@dvaldivia PTAL - esp. https://github.com/minio/kes/pull/111/files#diff-5cf40da47d04b063455c5695e527445a1ac43902d8b35ffa21aaba1aa23126c0R77

aead:
ping

Alevsk (approved):
Tested, works fine

This will require a change in console on the create tenant with encryption endpoint: https://github.com/minio/console/blob/master/pkg/kes/kes.go#L148 (rename kes to keyStore)

PR #120, opened 2021-04-21
add internal/fips package for FIPS 140 support

aead:
This commit adds a `internal/fips` package.
Now, the KES server / CLI can be built using
the Go BoringCrypto FIPS 140 module. This module
is only available on linux/amd64.

In FIPS mode, a KES server will only used NIST/FIPS
approved primitives and reject any ciphertexts (including
valid ones) generated e.g. using ChaCha20-Poly1305.

### DIFF
--- a/internal/fips/fips.go
+++ b/internal/fips/fips.go
@@ -1,0 +1,9 @@
+// Copyright 2021 - MinIO, Inc. All rights reserved.
+// Use of this source code is governed by the AGPLv3
+// license that can be found in the LICENSE file.
+
+// +build fips,linux,amd64
+
+package fips
+
+const enabled = 0 == 0
--- a/internal/fips/nofips.go
+++ b/internal/fips/nofips.go
@@ -1,0 +1,9 @@
+// Copyright 2021 - MinIO, Inc. All rights reserved.
+// Use of this source code is governed by the AGPLv3
+// license that can be found in the LICENSE file.
+
+// +build !fips
+
+package fips
+
+const enabled = 0 == 1
--- a/cmd/kes/main.go
+++ b/cmd/kes/main.go
@@ -15,7 +15,7 @@ import (
 	"strings"
 
 	"github.com/minio/kes"
-	"golang.org/x/crypto/ssh/terminal"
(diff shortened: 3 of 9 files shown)


harshavardhana (approved):

internal/fips/nofips.go
```diff
@@ -0,0 +1,9 @@
+// Copyright 2021 - MinIO, Inc. All rights reserved.
+// Use of this source code is governed by the AGPLv3
+// license that can be found in the LICENSE file.
+
+// +build !fips
+
+package fips
+
+const enabled = 0 == 1
```
donatello, line 9: Why use `0 == 1` instead of just `false`?

donatello (approved):
LGTM. Just have a small question.

internal/fips/fips.go
```diff
@@ -0,0 +1,9 @@
+// Copyright 2021 - MinIO, Inc. All rights reserved.
+// Use of this source code is governed by the AGPLv3
+// license that can be found in the LICENSE file.
+
+//+build fips,linux,amd64
```
harshavardhana, line 5:
```suggestion
// +build fips,linux,amd64
```

harshavardhana (changes requested):

internal/fips/nofips.go
aead: This is kind of an excessive measurement but in Go `true` and `false` are valid identifiers.
The following is valid Go:
```go
package main

import (
	"fmt"
)

const true = false

func main() {
	fmt.Println(true) // Oh no...
}
```

Therefore, the "recommendation" to make really sure that a boolean constant is evaluated by compiler like it you expect is to use a boolean expression - like `0 == 0`. More of a compliance thing / theoretical discussion xD

harshavardhana: @aead is trying to be extra careful since its KES :-p

donatello: Ah ok.

donatello (approved):

harshavardhana (approved):

PR #123, opened 2021-05-07
metric: add counters for error and audit events

aead:
This commit adds two metrics counting the number
of error resp. audit events.

Further, accessing the `/v1/metrics` API does no
longer produce an audit event. The `/v1/metrics`
was already excluded from the metrics collection.

It is also very unlikely that any audit log monitoring
is interested in the `/v1/metrics` API. A monitoring
system (e.g. Prometheus) will scrape the `/v1/metrics`
endpoint quiet often (every few seconds) leading to a
lot of audit event "noise".

Therefore, this PR disables audit events for the `/v1/metrics`
endpoint. If comprehensive audit events are desired this
can be enabled - as an option - in the future.

Now, the KES server metrics contains two additional counters:
```
 # HELP kes_log_audit_events Number of audit log events written to the audit log targets.
 # TYPE kes_log_audit_events counter
 kes_log_audit_events 2
 # HELP kes_log_error_events Number of error log events written to the error log targets.
 # TYPE kes_log_error_events counter
 kes_log_error_events 3
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -732,6 +732,8 @@ func (c *Client) Metrics(ctx context.Context) (Metric, error) {
 		MetricRequestErr    = "kes_http_request_error"
 		MetricRequestFail   = "kes_http_request_failure"
 		MetricRequestActive = "kes_http_request_active"
+		MetricAuditEvents   = "kes_log_audit_events"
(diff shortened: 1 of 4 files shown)


nitisht (approved):
LGTM

PR #128, opened 2021-05-24
policy: add deny rules

aead:
This commit adds `deny` rules - in addition to
the existing `allow` rules.

Before, a KES server policy contained a list of
patterns. If a request URL path matches any of
the pattern the request was accepted.

Now, a KES server policy can also contain a list
of `deny` patterns. If a request URL path matches
any of the `deny` patterns then the request is
rejected.

This is an API breaking change since the JSON representation
of a policy is changed by this commit. However, this
is acceptable since we have not committed to a stable `v1.0.0`
version, yet. From now on, more fields can be added to the KES
policy representation backwards-compatibly.

Before, a KES policy had the following format:
```
policy:
  my-policy:
    paths:
    - /v1/key/create/*
    - /v1/key/delete/*
    identities:
    - ...
```

From now on, a policy is defined using the following format:
```
policy:
  my-policy:
    allow:
    - /v1/key/create/*
    - /v1/key/delete/*
    deny:
    - /v1/key/delete/my-key
    identities:
    - ...
```

***

`Allow` and `deny` rules may overlap (one URL path may match
an `allow` and a `deny` pattern at the same time). A KES server
policy has the following semantics:
 1. If any deny rule matches the URL path then the request is rejected.
 2. Otherwise, if any allow rule matches the URL path the requested is
    accepted.
 3. Otherwise, the request is rejected.

So, a `deny` rule takes precedence over an `allow` rule. The main
usage of `deny` rules is to restrict a set of `allow` rules. For
example, an application may be allowed to decrypt DEKs generated
by `my-key*` except for `my-key-2` and `my-key-internal*`.

This can be expressed as following:
```
policy:
  my-policy:
    allow:
    - /v1/key/decrypt/my-key*
    deny:
    - /v1/key/decrypt/my-key-2
    - /v1/key/decrypt/my-key-internal*
    identities:
    - ...
```

This commit also temp. disables the policy integration tests since
this policy change is an API breaking change. Once we have released
this change (and once the CI instance has upgraded to it) we should
enable these tests again.

### DIFF
--- a/cmd/kes/config.go
+++ b/cmd/kes/config.go
@@ -42,10 +42,7 @@ type serverConfig struct {
 		} `yaml:"proxy"`
 	} `yaml:"tls"`
 
-	Policies map[string]struct {
-		Paths      []string       `yaml:"paths"`
-		Identities []kes.Identity `yaml:"identities"`
-	} `yaml:"policy"`
+	Policies map[string]policyConfig `yaml:"policy"`
 
 	Cache struct {
 		Expiry struct {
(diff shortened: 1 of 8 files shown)


Alevsk (approved):
LGTM

PR #129, opened 2021-05-26
use streaming API for listing identities

aead:
This commit modifies the `/v1/identity/list` API
to respond a nd-json stream of identity<->policy
mappings instead of one (eventually large) json
object.

The `/v1/identity/list` API is now similar to
the `/v1/key/list` API.

With the streaming API, a client can iterate over
some/all identities while using just a constant
amount of memory.

This is a major breaking API change.


Alevsk (approved):
LGTM

PR #131, opened 2021-06-18
add support for TLS 1.2

aead:
This commit adds support for TLS 1.2 in
addition to TLS 1.3.

Some legacy language versions (like Java 8)
do not support TLS 1.3, and therefore, fail
to connect to KES.

However, we want to support only PFS TLS connections.
Therefore, we limit the supported cipher suites to
ECDHE key exchange and AEAD ciphers.

Further, we have to drop ChaCha20-Poly1305 in case
of FIPS 140.

### DIFF
--- a/cmd/kes/server.go
+++ b/cmd/kes/server.go
@@ -27,6 +27,7 @@ import (
 	"github.com/fatih/color"
 	"github.com/minio/kes"
 	"github.com/minio/kes/internal/auth"
+	"github.com/minio/kes/internal/fips"
 	xhttp "github.com/minio/kes/internal/http"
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

harshavardhana:
> Some legacy language versions (like Java 8)

Java8 is EOL'ed for most apps by Oracle https://www.infoworld.com/article/3532358/oracle-extends-extended-support-for-java-8.html 

https://www.infoworld.com/article/3269332/oracle-sets-date-for-end-of-java-8-updates.html

Using this can have unforeseen situations in the future. 

Without commercial support from Oracle Java8 is EOL'ed and shouldn't be used.

PR #132, opened 2021-06-18
fix: release kes containers to quay.io as well

harshavardhana:


internal/http/handler.go
```diff
@@ -358,17 +358,18 @@ func HandleDecryptKey(store *secret.Store) http.HandlerFunc {
 		}
 		secret, err := store.Get(name)
 		if err != nil {
-			Error(w, err)
+			Error(w, fmt.Errorf("store.Get: %w", err))
```
aead, line 361: We should not wrap errors here:
 - We should not disclose to clients what specific operation failed. Especially not for decryption since we create an oracle for the client: https://en.wikipedia.org/wiki/Oracle_attack
 - This will mask errors clients may want to check. For example: https://github.com/minio/kes/blob/master/internal/vault/vault.go#L215

internal/http/handler.go
```diff
@@ -358,17 +358,18 @@ func HandleDecryptKey(store *secret.Store) http.HandlerFunc {
...
 			return
 		}
 		plaintext, err := secret.Unwrap(req.Ciphertext, req.Context)
 		if err != nil {
-			Error(w, err)
+			Error(w, fmt.Errorf("secret.Unwrap: %w", err))
```
aead, line 366: Same as above.

aead (changes requested):

internal/http/handler.go
aead: If such information is really important and indicates a actionable error we should rather create a error log entry.

aead (approved):

PR #134, opened 2021-06-28
add support for Azure KeyVault key store

aead:
This commit adds support for Azure KeyVault
as key store.

The current implementation only supports
client credentials, not Azure machine identities.
Support for machine identities is planned for
the future.

However, Azure KeyVault does not offer atomic
create or delete semantics. Therefore, concurrent
stateful operations - e.g. creating a key - can
lead to data race situations.

KES mitigates the effects of such data races using
key IDs which are included into generated ciphertexts
as well as defensive code that tries to reduce the
race condition window as much as possibles.


harshavardhana:
> It has to be documented that in a HA setup / setup
> with multiple KES servers data races can occur when
> stateful operations are executed concurrently. In the
> worst case, these data races can lead to data corruption
> or data loss.

I am a bit worried about this, can we not use some kind of lease/locking mechanism on the azure key vault?

Alevsk (approved):
LGTM we will have to bring this functionality to operator-ui security configuration as well

aead:
#141 Is a precondition for this PR. It will address and try to mitigate some of the data race effects.

Alevsk (approved):
LGTM

harshavardhana (approved):

PR #136, opened 2021-07-06
vault: add support for the K/V v2 engine

aead:
This commit adds support for the K/V v2 engine.

The K/V v2 engine, in contrast to the v1 engine,
supports versioned secrets. However, KES requires
atomic stateful operations (e.g. create or delete a key)
to avoid data races that can lead to data corruption
and data loss for KES clients.

Therefore, KES utilizes the CAS option when creating keys.
It ensures that an entry can only be created when no entry
exists already. If this option would not be set, KES would
eventually overwrite an existing key / create a new version
in distributed deployments.

Further, KES uses the atomic delete all versions and metadata
API - instead of deleting the key and its metadata separately.

***

The K/V v2 engine also requires different policy permissions
compared to K/V v1. For a comprehensive explanation visit:
https://www.vaultproject.io/docs/secrets/kv/kv-v2#acl-rules

To summarize:
K/V v2 distinguishes between `data` and `metadata` operations.
Therefore, some API operations - like `Create` require the
`data` permission and some - like `List` require the metadata
permission.

The following HCL policy grants KES full read/write/list access
on the K/V v2 engine:

```
path "kv/data/*" {
     capabilities = [ "create", "read" ]
}
path "kv/metadata/*" {
     capabilities = [ "list", "delete" ]
}
```


cmd/kes/config.go
```diff
@@ -329,6 +331,9 @@ func (config *kmsServerConfig) SetDefaults() {
 	if config.Vault.EnginePath == "" {
 		config.Vault.EnginePath = "kv" // If not set, use the Vault default engine path.
 	}
+	if config.Vault.EngineVersion == "" {
+		config.Vault.EngineVersion = "v1"
```
harshavardhana, line 335: Can this be a constant like `engineV1` ?

harshavardhana (changes requested):

harshavardhana (approved):

PR #141, opened 2021-08-03
secret: add key ID to encrypted keys

aead:
This commit adds unique key IDs to generated
ciphertexts. The key ID identifies the root/master
key used to generate the ciphertext. Now it is possible,
to determine whether a particular key was used to encrypt
a certain ciphertext.

However, the key ID is just for informational purposes.
It is derived from the root/master key content as:
```
id := HEX(SHA-256(key)[:16])
```

The ID is a truncated hash for two reasons:
 1. Avoid additional size overhead
 2. Limit the damage a broken hash function (i.e. SHA-256)
    could cause.

W.r.t. truncating SHA-256 to 128 bits:
This is not a security issue but instead a common technique.
In particular, SHA-224 is truncated SHA-256 and SHA-384 is truncated
SHA-512.
Truncating the hash only reduces the collision resistance boundary.
However, we are not interested in collision resistance w.r.t. an
attacker crafting a collision. Instead, we only want to use that
IDs are "sufficiently" unique - such that two keys have different
IDs with very high probability.

The most important property of the ID is that it must not reveal
anything about the root/master key content.

This should be the case since we use a collision resistant hash
function. Further, even if SHA-256 would reveal information about
the hashed message (and therefore would be crypto. broken) we
limit the damage by truncating the hash to 128 bits. It's very
unlikely that a 128 bit hash value would reveal more than 128 bit
about the message - here the hashed key.

### DIFF
--- a/internal/secret/store.go
+++ b/internal/secret/store.go
@@ -115,7 +115,6 @@ func (s *Store) Create(name string, secret Secret) (err error) {
 	if err = s.Remote.Create(name, secret.String()); err != nil {
 		return err
 	}
-	s.cache.SetOrGet(name, secret)
 	return nil
 }
 
@@ -145,7 +144,7 @@ func (s *Store) Get(name string) (Secret, error) {
 	if err != nil {
 		return Secret{}, err
 	}
-	return s.cache.SetOrGet(name, secret), nil
+	return s.cache.CompareAndSwap(name, secret), nil
 }
 
 // List returns a new Iterator over all key names.
--- a/internal/secret/cache.go
+++ b/internal/secret/cache.go
@@ -43,16 +43,15 @@ func (c *cache) Set(name string, secret Secret) {
 	}
 }
 
-// SetOrGet adds  given secret to the cache
-// if and only if no entry for name already
-// exists. Instead, if an entry for the given
-// name exists it returns the secret that is
-// currently present.
+// CompareAndSwap adds the secret to the cache
+// if and only if no entry for given name exists.
+// If an entry for the given name exists it returns
+// the secret that is currently in the cache.
 //
-// SetOrGet will always return the secret that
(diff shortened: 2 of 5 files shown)


internal/secret/store.go
```diff
@@ -115,7 +115,6 @@ func (s *Store) Create(name string, secret Secret) (err error) {
 	if err = s.Remote.Create(name, secret.String()); err != nil {
 		return err
 	}
-	s.cache.SetOrGet(name, secret)
```
harshavardhana, line 118: Is this expected? @aead

harshavardhana (changes requested):

internal/secret/store.go
aead: Yes @harshavardhana - this another improvement against data races. Now, creating a key does not load the created key directly into the cache. Instead, KES fetches the key on it's fist actual usage.

That means, that two concurrent `CreateKey` requests to two KES servers don't affect the caches of those KES servers. Instead, they fetch the created key once it is actually used. Now, the likelihood of two KES servers holding with different keys (with the same name) in the cache becomes significantly smaller.

harshavardhana: I see @aead is there a likely hood of this happening with kes v0.10 versions

Alevsk (approved):

internal/secret/store.go
aead: They can - but all all other backends, except for Azure KeyVault, support atomic creates. So, if two `CreateKey` calls happen concurrently then one will win while the other will fail. Therefore, it was "cheaper" to also load the key into the cache once creation succeeded. However, this was a micro optimisation and should not really matter w.r.t. performance.

aead: The data race issues / concurrency problems are specific to Azure KeyVault.

PR #142, opened 2021-08-16
http: fix type conversion in timeout handler

aead:
This commit fixes an incorrect type conversion
in the http timeout handler implementation.

Code like `x := y.(Interface)` leads to a panic
at runtime. Instead, you have to check the type
conversion:
```
var x Interface
if t, ok := y.(Interface); ok {
   x = t
}
```


harshavardhana (approved):

vadmeste (approved):
LGTM

PR #143, opened 2021-08-18
api: new status API

aead:
This commit adds a new status API:
```
https://<endpoint>/v1/status
```

The status API exposes some status information
about the server (up time, version, TLS certificate
expiry, ...).

As part of this API, KES now reloads the TLS certificate
periodically such that changing the server certificate/private key
will be reflected eventually.


internal/http/handler.go
```diff
@@ -120,6 +121,41 @@ func HandleVersion(version string) http.HandlerFunc {
...
+		Certificate struct {
+			Expiry   time.Duration `json:"expiry"`
+			DNSNames []string      `json:"dns,omitempty"`
+			IPs      []net.IP      `json:"ip,omitempty"`
+		} `json:"certificate"`
+	}
+	var startTime = time.Now()
+	return func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
```
harshavardhana, line 144:
```suggestion
		w.WriteHeader(http.StatusOK)
```

This is not needed if you are ever going to always send success to client @aead - can we not have provisions for this endpoint to return error at all ?

harshavardhana (changes requested):

internal/http/handler.go
aead: > This is not needed if you are ever going to always send success to client @aead
Correct, not strictly needed when you at least call `Write` once.

> can we not have provisions for this endpoint to return error at all ?
You mean: "KES starts but is not able to perform all operations - like MinIO server starting but some features are not available?"

No, this cannot happen. Either KES starts and is fully functional or KES does not start and fails with an error explaining why.

harshavardhana: Okay

harshavardhana (approved):

internal/http/handler.go
```diff
@@ -120,6 +121,41 @@ func HandleVersion(version string) http.HandlerFunc {
...
+
+		var status = Status{
+			Version: version,
+			UpTime:  time.Since(startTime).Truncate(time.Second),
+		}
+		if cert, err := certificate.GetCertificate(nil); cert != nil && err == nil {
+			status.Certificate.Expiry = time.Until(cert.Leaf.NotAfter).Truncate(time.Second)
+			status.Certificate.DNSNames = cert.Leaf.DNSNames
+			status.Certificate.IPs = cert.Leaf.IPAddresses
```
harshavardhana, line 153: On second thought should we log an error here? or send it to client?

harshavardhana (changes requested):

harshavardhana (dismissed):

internal/http/handler.go
aead: This should not happen since a) KES requires TLS and b) `GetCertificate` should always return a certificate.
However, we don't provide a `tls.ClientHello` (but `nil` since we don't have one) and `GetCertificate` may do things like SNI certificate mapping.  
If `certificate.GetCertificate(nil)` returns `nil` / an error then `GetCertificate` sporadically fails to return a certificate. It has to succeed at least sometimes - otherwise, the client could not even connect to the server. 

However, we could log an error message. Sending an error to the client may be odd since from client perspective it will look tike:
 1. Client to Server connection succeeded. Therefore `GetCertificate` returned a certificate
 2. Then `GetCertificate` failed and the server returned an error.
 So, for the client it would look like the server first sends its TLS certificate and then the server claims to not have a TLS certificate.

aead: Added error logging for this case.

harshavardhana (approved):

PR #144, opened 2021-08-27
vault: ensure that a write to Vault actually succeeded

aead:
This commit fixes a subtle issue caused by questionable
behavior of the Vault SDK.

When writing an entry to Vault's K/V engine, the Vault
SDK returns an object representing the entry or an error
explaining why the write failed.

However, the Vault SDK may return no entry object and no
error even if writing to Vault failed - e.g. in case of
a broken network connection.

This can cause KES to think that it created a key successfully
while Vault actually didn't receive the request. So, KES may
return 200 OK to its client while the key creation actually failed.

This commit fixes this by checking that we actually got an
entry object from Vault.

### DIFF
--- a/internal/vault/vault.go
+++ b/internal/vault/vault.go
@@ -368,9 +368,18 @@ func (s *Store) Create(key, value string) error {
 			key: value,
 		}
 	}
-	_, err := s.client.Logical().Write(location, data)
+
+	// The Vault SDK may return no error even if it hasn't created
+	// an entry - e.g. in case of some network errors. Therefore,
+	// we also check that the returned entry is not nil to ensure
+	// that we got a response from the Vault cluster.
+	entry, err := s.client.Logical().Write(location, data)
 	if err != nil {
 		s.logf("vault: failed to create '%s': %v", location, err)
+		return errCreateKey
+	}
+	if entry == nil {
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

harshavardhana (commented):
```go
func (c *Logical) write(path string, request *Request) (*Secret, error) {
	ctx, cancelFunc := context.WithCancel(context.Background())
	defer cancelFunc()
	resp, err := c.c.RawRequestWithContext(ctx, request)
	if resp != nil {
		defer resp.Body.Close()
	}
	if resp != nil && resp.StatusCode == 404 {
		secret, parseErr := ParseSecret(resp.Body)
		switch parseErr {
		case nil:
		case io.EOF:
			return nil, nil
```

```go

// ParseSecret is used to parse a secret value from JSON from an io.Reader.
func ParseSecret(r io.Reader) (*Secret, error) {
	// First read the data into a buffer. Not super efficient but we want to
	// know if we actually have a body or not.
	var buf bytes.Buffer
	_, err := buf.ReadFrom(r)
	if err != nil {
		return nil, err
	}
	if buf.Len() == 0 {
		return nil, nil
	}
```

🔫

PR #147, opened 2021-09-02
key: refactor key store backend API

aead:
This commit refactors the key store API
to use a new `key` package as a replacement
for the `secret` package.

The new `key` package is a prerequisite for
further improvements for keys - i.e. typed
keys, sym. vs. asym. keys etc.

Further, this commit cleans up some existing
code and defines the responsibilities of different
types more precisely - i.e. `key.Manager` vs. `key.Store`.

This commit should not cause any change in behavior.


Alevsk (approved):
lgtm

PR #151, opened 2021-09-17
fortanix: add support for Fortanix SDKMS

aead:
This commit adds a Fortanix SDKMS keystore
backend. Now, Fortanix SDKMS can be used
to store and fetch keys.

This commit only implements a single authentication
method: API keys. It does not support other authentication
methods like JWT, LDAP, Google Service Accounts, ...

A minimal KES<--->Fortanix configuration may be:
```yaml
root: 3ecfcdf38fcbe141ae26a1030f81e96b753365a46760ae6b578698a97c59fd22

tls:
  key: private.key
  cert: public.crt

keystore:
  fortanix:
    sdkms:
      endpoint: "https://sdkms.fortanix.com"
      credentials:
        key: "NWMyMWZlNzktZDRmZS00NDFhLWFjMzMtNjZmY2U0Y2ViMThhOnVpUElyWUlhc2FIdkExLVRUOXJFa0ZpSnVPaVpEQTFKZ2hzYUZGT29mcmd3UEJwUnZmcmM0REJzd2UwTk5vSF9oU3BzTGlJN0o2Z3JUTEFPaVR0bV9B"
```

### DIFF
--- a/internal/fortanix/keystore.go
+++ b/internal/fortanix/keystore.go
@@ -1,0 +1,639 @@
+// Copyright 2021 - MinIO, Inc. All rights reserved.
+// Use of this source code is governed by the AGPLv3
+// license that can be found in the LICENSE file.
+
+package fortanix
+
+import (
+	"bytes"
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"net"
+	"net/http"
+	"net/url"
+	"os"
+	"path"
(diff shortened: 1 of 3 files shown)


internal/fortanix/keystore.go
```diff
@@ -0,0 +1,635 @@
+package fortanix
```
harshavardhana, line 1: License header @aead

harshavardhana (changes requested):

internal/fortanix/keystore.go
aead: Oh, thanks 👍 - fixed!

harshavardhana (approved):

PR #157, opened 2021-09-29
gateway: introduce gateway command

aead:
This commit introduces the `kes gateway`
command. The `gateway` command currently
behaves exactly as the `kes server` command.
    
In the future, `kes server` will implement
a stateful KES server that persists keys itself.
    
KES infront of another KMS (Hashicorp Vault, AWS, ...)
will become implemented `kes gateway` - acting as a
gateway infront of the central KMS.


aead:
Depends on #156

PR #160, opened 2021-10-11
config: rename root identity

aead:
This commit renames the root identity to
admin identity. Therefore, it changes
the config file structure:

Before:
```yaml

root: ...
```

After:
```yaml

admin:
  identity: ...
```

Further, this commit removes the `--root` CLI flag.
From now on, the admin identity has to be specified
in the config file - but it can be inferred from the
environment.

Signed-off-by: Andreas Auernhammer <hi@aead.dev>

### DIFF
--- a/cmd/kes/config.go
+++ b/cmd/kes/config.go
@@ -30,8 +30,10 @@ import (
 )
 
 type serverConfig struct {
-	Addr string       `yaml:"address"`
-	Root kes.Identity `yaml:"root"`
+	Addr  string `yaml:"address"`
+	Admin struct {
+		Identity kes.Identity `yaml:"identity"`
+	} `yaml:"admin"`
 
 	TLS struct {
 		KeyPath  string `yaml:"key"`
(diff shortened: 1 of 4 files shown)


bboreham:
May I suggest updating the instructions at https://github.com/minio/kes/wiki/Getting-Started which don't work since `--root` has been removed.

harshavardhana:
> May I suggest updating the instructions at https://github.com/minio/kes/wiki/Getting-Started which don't work since `--root` has been removed.

@aead isn't this is expected to be backward compatible?

aead:
@bboreham Thanks for the hint. We will update the guides accordingly. However, this change is not released yet. It is only removed in upstream/master In the future, we will expect a config file.

@harshavardhana We will drop the `--root` (and probably other flags) to simply the CLI. Maybe we will need new flags for the stateful cluster mode.

bboreham:
It was released as part of https://github.com/minio/kes/releases/tag/v0.17.2

PR #171, opened 2021-12-07
status: expose backend KMS status

aead:
This commit changes the `/v1/status` API by:
 - adding a backend KMS status
 - removing the TLS certificate information

The KMS status is generic for all KMS backends
and describes whether the backend KMS is reachable
or unreachable as well as the network latency between
KES and the backend KMS instance.

The TLS certificate information got removed from the
status API since this information can be gathered by
the client itself (by looking at the TLS certificate).

Exposing this information in the status API is particularly
useful and just bloats the API.

### DIFF
--- a/internal/http/handler.go
+++ b/internal/http/handler.go
@@ -8,7 +8,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
-	"net"
 	"net/http"
 	"path"
 	"strconv"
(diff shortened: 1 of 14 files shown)


internal/http/handler.go
```diff
@@ -127,35 +126,31 @@ func handleVersion(version string) http.HandlerFunc {
...
+			Latency time.Duration `json:"latency,omitempty"`
+		} `json:"kms"`
 	}
 	var startTime = time.Now()
 	return func(w http.ResponseWriter, r *http.Request) {
-		w.Header().Set("Content-Type", "application/json")
+		kmsState, err := manager.Store.Status(r.Context())
+		if err != nil {
+			log.Log().Printf("http: failed to connect to key store: %v", err)
+		}
```
harshavardhana, line 144: Is it not expected to return from this handler after error?

internal/key/store.go
```diff
@@ -60,3 +68,80 @@ type Iterator interface {
...
+	)
+	if port == "" {
+		port = DefaultPort
+	}
+
+	var (
+		d     net.Dialer
+		start = time.Now()
+	)
+	c, err := d.DialContext(ctx, "tcp", host+":"+port)
```
harshavardhana, line 130:
```suggestion
	c, err := d.DialContext(ctx, "tcp", net.JoinHostPort(host, port))
```

harshavardhana: I think it's better to also reach out to https://www.vaultproject.io/api/system/health for Vault-specific endpoints to health. 

Generic Dial is okay and serves it purpose but we should consider writing health checks for individual targets.

harshavardhana (changes requested):

internal/http/handler.go
aead: We don't want to abort the `status` API call "just" because fetching the KMS status failed.
We return an `unreachable` KMS status instead and just log that we failed to fetch the KMS status.

internal/key/store.go
aead: > Generic Dial is okay and serves it purpose but we should consider writing health checks for individual targets.

Yes, that we can do. I added a health check for Vault. Others can follow. However, I don't want to expose target-specific details in the API. For example, AWS is not in a sealed or not initialized state - in contrast to Vault.

For a client it is sufficient to distinguish:
 - `available`: The KMS is reachable and ready to serve requests
 - `reachable`: The KMS is reachable but is may not be able to serve requests
 - `unreachable`: The KMS is not reachable (over the network)

internal/http/handler.go
harshavardhana: Okay

kannappanr (approved):
LGTM

PR #172, opened 2021-12-09
cache: add offline caching of keys 

aead:
This commit adds offline caching of keys.

KES keeps keys - fetched from the KMS key
store - in a in-memory cache to increase
performance and reduce the request rate
to the central KMS.
Entries in the cache expiry after a configurable
time period and get removed by a cache GC.

Usually it is recommended to keep the cache
expiry periods quite low - e.g.:
 - Any:    5m0s
 - Unused:  20s

In particular, low cache expiry values reduce
the time window KES can operate without interacting
with the central KMS when serving stateless
requests; e.g. generating a new data encryption key.
Note that KES can never server stateful requests,
like creating or deleting a key, without the KMS.

Especially in distributed setups, two KES servers
will sync eventually once their cache entires have
expired. For example, one KES server receives a
request to delete a key from the KMS key store.
The second KES server will not notice that this key
got deleted until its corresponding cache entry
has expired.
Low cache expiry values reduce the time window
when multiple KES servers are not synchronized.

However, low cache expiry values require that the
KMS is highly available. As soon as a cache entry
expires, KES needs to reach out to the KMS to
fetch the key again.
If the KMS is not available, KES will not be able
continue serving stateless requests.

It may be desirable to keep keys longer in the cache
to reduce the impact of the central KMS being down
and continue serving stateless requests - but only
when the KMS is actually down. When the KMS is
available, KES should expiry keys relatively quickly
and only cache them longer when the KMS is not reachable.

This commit adds this ability by another cache expiry
configuration:
```yaml
cache:
  expiry:
    any:       5m0s
    unused:     30s
    offline: 1h0m0s
```

Now, KES will cache keys for one hour if and only
if the KMS is not available. As soon as the KMS
is reachable again, KES clears the cache to sync
with the central KMS again.

If no `offline` expiry is set, KES will not cache
keys when the KMS is down. It will simply not
use an offline cache.

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -41,12 +41,11 @@ var _ key.Store = (*Store)(nil)
 
 // Status returns the current state of the FS key store.
 func (s *Store) Status(_ context.Context) (key.StoreState, error) {
-	var (
-		start   = time.Now()
-		_, err  = os.Stat(s.Dir)
-		latency = time.Since(start)
-		state   = key.StoreAvailable
-	)
+	state := key.StoreAvailable
+
+	start := time.Now()
+	_, err := os.Stat(s.Dir)
+	latency := time.Since(start)
 	if err != nil {
 		state = key.StoreUnreachable
 	}
--- a/internal/key/store.go
+++ b/internal/key/store.go
@@ -3,6 +3,7 @@ package key
 import (
 	"context"
 	"errors"
+	"log"
 	"net"
 	"net/url"
 	"time"
@@ -13,6 +14,16 @@ import (
 type Store interface {
 	// Status returns the current state of the
 	// Store.
+	//
+	// If Status fails to reach the Store - e.g.
+	// due to a network error - it should return
+	// a StoreState with StoreUnreachable and no
+	// error.
+	//
+	// Status should return an error whenever it
+	// fails to reach the Store but StoreUnreachable
+	// is not appropriate to describe the error
+	// condition.
 	Status(context.Context) (StoreState, error)
 
 	// Create stores the given key at the key store if
@@ -166,4 +177,34 @@ func DialStore(ctx context.Context, endpoint string) (StoreState, error) {
 		State:   StoreReachable,
 		Latency: latency,
 	}, nil
+}
+
+// LogStoreStatus periodically fetches the Store status
+// and writes a log message whenever the Store is not
+// available.
+//
+// It stops whenever the given Context.Done() channel
+// returns.
+func LogStoreStatus(ctx context.Context, store Store, interval time.Duration, out *log.Logger) {
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	var lastAvailable time.Time
+	for {
+		state, err := store.Status(ctx)
+		switch {
+		case err != nil:
+			out.Printf("key: failed to fetch KMS key store status information: %v", err)
+		case state.State != StoreAvailable:
+			out.Printf("key: KMS key store is %s for %v", state.State, time.Since(lastAvailable).Round(time.Second))
+		default:
+			lastAvailable = time.Now().UTC()
+		}
+
+		select {
+		case <-ticker.C:
+		case <-ctx.Done():
+			return
+		}
+	}
 }
--- a/cmd/kes/server.go
+++ b/cmd/kes/server.go
@@ -207,9 +207,10 @@ func server(args []string) {
 		stdlog.Fatalf("Error: %v", err)
 	}
 	manager := &key.Manager{
-		CacheExpiryAny:    config.Cache.Expiry.Any.Value(),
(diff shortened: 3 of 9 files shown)


aead:
Depends on #171

internal/fs/fs.go
```diff
@@ -38,6 +39,23 @@ type Store struct {
...
 var _ key.Store = (*Store)(nil)
 
+// Status returns the current state of the FS key store.
+func (s *Store) Status(_ context.Context) (key.StoreState, error) {
+	var (
+		start   = time.Now()
+		_, err  = os.Stat(s.Dir)
+		latency = time.Since(start)
+		state   = key.StoreAvailable
+	)
```
klauspost, line 49: I bit clearer IMO:
```suggestion
	state := key.StoreAvailable
	start := time.Now()
	_, err := os.Stat(s.Dir)
	latency := time.Since(start)
```

internal/key/store.go
```diff
@@ -2,11 +2,19 @@ package key
...
+	"net/url"
+	"time"
 )
 
 // Store is a key store that persists keys that
 // are referenced by a unique name.
 type Store interface {
+	// Status returns the current state of the
+	// Store.
+	Status(context.Context) (StoreState, error)
```
klauspost, line 16: Document how to interpret a returned error compared to "StoreUnreachable" state.

internal/vault/vault.go
```diff
@@ -129,6 +129,54 @@ var (
...
+	// but the sys/health API defaults to returning 5xx when not sealed or
+	// initialized, so we force this code to be something else so we parse correctly
+	req.Params.Add("uninitcode", "299")
+	req.Params.Add("sealedcode", "299")
+	req.Params.Add("standbycode", "299")
+	req.Params.Add("drsecondarycode", "299")
+	req.Params.Add("performancestandbycode", "299")
+
+	resp, err := s.client.RawRequestWithContext(ctx, req)
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
```
klauspost, line 159: Should state be modified if we cannot make a health check?

klauspost (commented):
Just a few driveby comments.

internal/vault/vault.go
aead: In this case `state.State == StoreReachable` (since we were able to dial Vault but failed to check Vault's health).
This seems to be appropriate to describe the current situation. Vault is reachable but (probably not available).

harshavardhana (changes requested):
@aead should we log something on the kes console regarding when the server moves to 'offline' mode? and switches back? 

This allows us to track how unstable is the backend. If we also capture this in metrics is useful for us to help customers and we can also point to this as debugging information for us.

harshavardhana:
> @aead should we log something on the kes console regarding when the server moves to 'offline' mode? and switches back?
> 

For now just a log is sufficient

> This allows us to track how unstable is the backend. If we also capture this in metrics is useful for us to help customers and we can also point to this as debugging information for us.

This is good to have but not sure it will be easily possible to be exposed or consumed.

klauspost (approved):
Traced call paths. Locking seems solid.

aead:
@harshavardhana Added KMS key store monitoring that keeps logging once the KMS key store becomes unavailable until it recovers. For example:

```
2021/12/17 13:45:24 key: KMS keystore is unreachable since 1m0s
2021/12/17 13:46:24 key: KMS keystore is unreachable since 2m0s
2021/12/17 13:47:24 key: KMS keystore is unreachable since 3m0s
```

Once, the KMS key store is available again we stop logging.
We can expose this as metrics but let's leaf that to a separate PR.

internal/key/store.go
```diff
@@ -167,3 +178,32 @@ func DialStore(ctx context.Context, endpoint string) (StoreState, error) {
...
+// returns.
+func LogStoreStatus(ctx context.Context, store Store, interval time.Duration, out *log.Logger) {
+	var lastAvailable time.Time
+	for {
+		state, err := store.Status(ctx)
+		switch {
+		case err != nil:
+			out.Printf("key: failed to fetch KMS key store status information: %v", err)
+		case state.State != StoreAvailable:
+			out.Printf("key: KMS key store is %s since %s", state.State, time.Since(lastAvailable).Round(time.Second))
```
klauspost, line 196:
```suggestion
			out.Printf("key: KMS key store is %s for %v", state.State, time.Since(lastAvailable).Round(time.Second))
```

internal/key/store.go
```diff
@@ -167,3 +178,32 @@ func DialStore(ctx context.Context, endpoint string) (StoreState, error) {
...
+		default:
+			lastAvailable = time.Now().UTC()
+		}
+
+		var timer = time.NewTimer(interval)
```
klauspost, line 201: nit: Use a single time.Ticker instead?

harshavardhana (approved):

poornas (approved):

PR #174, opened 2021-12-16
cli: use correct policy name instead of command name

aead:
This commit fixes a bug in the `kes policy add` command.
Before, the command would create a policy with the name
`add` instead of whatever policy name has been specified.


harshavardhana (approved):

harshavardhana:
@aead does this require a new release?

PR #179, opened 2022-01-04
vault: include K/V prefix when cloning config

### LINKED ISSUE #178: `prefix` does not take effect for Vault Keystore configuration
#### Bug description
For version 0.17.4, in `kes.yml`

```
...
keystore:
  vault:
    endpoint: http://vault.internal:8200
    prefix: staging
    approle:
      engine: staging/approle
      id: ""
      secret: ""
      retry: 15s
    status:
      ping: 10s
...
```

#### Expected behavior
Expected kes to fire to `http://vault.internal:8200/v1/kv/staging/key-name`, but it fired to `http://vault.internal:8200/v1/kv/key-name` instead, ignoring what is set in the `prefix` key.

#### Additional context

1. What version of Go are you using (`go version`)?

`go version go1.15.14 linux/amd64`

`RHEL 7.9 amd64`

We downgraded from v0.17.4 to v0.17.2 and it's working again. We are not proficient with Go but we combed through the code and it seems very likely that this commit [a229d68](https://github.com/minio/kes/commit/a229d68fd98192ab4df32c296ab5febec172f5b8) broke the feature, especially [here](https://github.com/minio/kes/blob/master/internal/vault/config.go#L154) where we saw that `prefix` are not being returned in the cloned object.

aead:
This commit fixes a bug introduced by a229d68
that ignores the K/V prefix when cloning the
Vault config.

This commit fixes this bug and adds a test to
ensure that a config is cloned as expected.

Fixes #178

### DIFF
--- a/internal/vault/config.go
+++ b/internal/vault/config.go
@@ -163,6 +163,7 @@ func (c *Config) Clone() *Config {
 		Engine:          c.Engine,
 		APIVersion:      c.APIVersion,
 		Namespace:       c.Namespace,
+		Prefix:          c.Prefix,
 		AppRole:         c.AppRole,
 		K8S:             c.K8S,
 		ErrorLog:        c.ErrorLog,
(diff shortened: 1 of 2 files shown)


klauspost (approved):
LGTM. Quick release coming?

aead:
I guess yes, `v0.17.5`

PR #181, opened 2022-01-06
yml: fix regression parsing policy definitions

### LINKED ISSUE #180: Policies not working after upgrading to v0.17.5
#### Bug description
For version 0.17.5, in `kes.yml`

```
...
policy:
  staging-minio:
    allow:
    - /v1/key/create/key-name
    - /v1/key/generate/key-name
    - /v1/key/decrypt/key-name
    identities:
    - value-of-minio-certificate
...
```

#### Expected behavior
While trying to authenticate a client using a policy, we are getting "prohibited by policy" in audit logs after upgrading to v0.17.5 without any changes to our `kes.yml`. We've checked and it seems that the policy is correct according to the documentation, and also that the correct identity was being used according to the audit logs.

#### Additional context

1. What version of Go are you using (`go version`)?

`go version go1.15.14 linux/amd64`

`RHEL 7.9 amd64`

We've downgraded from v0.17.5 to v0.17.2 and it's working again. We are wondering if [this](https://github.com/minio/kes/blob/a229d68fd98192ab4df32c296ab5febec172f5b8/internal/yml/server-config_v0.17.0.go#L25) should be `yaml:identities` instead of `yaml:identity` since the key name in the yaml should be `identities`

aead:
This commit fixes a regression introduced by
a229d68.

Identities assigned to a policy in the YAML
config file have been ignored in the following
case:
```
policy:
   - my-policy:
       allow:
         - /v1/key/create/*
       identity:
         - ${MY_CLIENT_IDENTITY}
```

Before, the config file parsing expected
`identities` - not `identity`. After a229d68
`identities` has been ignored. This commit
fixes this by reverting to the previous behavior
of only honoring `identities`.

Fixes #180

### DIFF
--- a/internal/yml/server-config.go
+++ b/internal/yml/server-config.go
@@ -31,7 +31,7 @@ type ServerConfig struct {
 	Policies map[string]struct {
 		Allow      []string   `yaml:"allow"` // Use 'string' type; We don't replace API allow patterns with env. vars
 		Deny       []string   `yaml:"deny"`  // Use 'string' type; We don't replace API deny patterns with env. vars
-		Identities []Identity `yaml:"identity"`
+		Identities []Identity `yaml:"identities"`
 	} `yaml:"policy"`
 
 	Cache struct {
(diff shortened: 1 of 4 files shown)


klauspost (approved):
LGTM

harshavardhana (approved):

PR #182, opened 2022-01-12
key: remove Manager and export Cache

aead:
This commit is a refactoring of the
`internal/key` package that removes
the `Manager` type and exports the
`Cache` type.

Now, an arbitrary key store can be
wrapped by a cache like this:

```
cache := key.NewCache(store, key.CacheConfig{
   Expiry:        5 * time.Minute,
   ExpiryUnused: 30 * time.Second,
})
```

Furthermore, a `Cache` satisfies the
`key.Store` interface. This is an important
property that will be relevant later on
when we add a stateful implementation.


klauspost (approved):
👍🏼

PR #183, opened 2022-01-19
sys: add minimal multi-tenancy support via Enclaves

aead:
This commit adds the most minimal multi-tenancy
implementation.

Now, KES contains an internal Vault. The Vault
has one or multiple Enclaves. (For now just one).

Each Enclave has its own key store, policies set
and identity set. A client selects the enclave
using the `?enclave=<name>` query parameter.

For example:
```
https://127.0.0.1:7373/v1/key/create/my-key?enclave=tenant-1
```

However, a stateless KES server will only support a
single Enclave. Therefore, this commit only adds the
most basic Vault implementation.

Future Vault implementations (e.g. a standalone FS
KES server) will use a more sophisticated Vault
implementation that supports seal/unsealing etc.

This commit mainly focuses on changing the current
code base to use a `sys.Vault` and don't change any
client-facing behavior.

### DIFF
--- a/cmd/kes/config.go
+++ b/cmd/kes/config.go
@@ -9,11 +9,14 @@ import (
 	"errors"
 	"fmt"
 	stdlog "log"
+	"net/http"
 	"os"
 	"path/filepath"
-	"strings"
+	"sync"
 	"time"
 
+	"github.com/minio/kes"
+	"github.com/minio/kes/internal/auth"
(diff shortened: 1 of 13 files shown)


Alevsk (approved):
+1 for using iterators, beautiful and well documented code :+1:

harshavardhana (approved):

PR #186, opened 2022-01-27
remove support for go1.16, enable gofumpt

harshavardhana:


klauspost:
Seems like a semi-legit complaint.

harshavardhana:
@klauspost validatePath code seems to handle that anyways key names do not support path segments - GitHub's codeQL doesn't understand or handle this part yet.

So it is a sort of false positive.

klauspost:
@harshavardhana `..` could still be sent as name AFAICT. Maybe even backslash on Windows. It would fail on it, but a bit stricter validation would probably be better - that is it is "semi-legit".

aead:
> .. could still be sent as name AFAICT. Maybe even backslash on Windows. It would fail on it, but a bit stricter validation would probably be better - that is it is "semi-legit".

Currently it cannot pass path validation. However, validation will become stricter very soon. I will restrict to very small character set when reworking / extending the API (stateful server).

aead (approved):

PR #189, opened 2022-02-03
kes: add Status API to client SDK

aead:
This commit adds the `Status` API call
to the client SDK.

The current client API only contains a
subset of the status information exposed
by the server due to API instability.


client.go
```diff
@@ -309,6 +309,29 @@ func (c *Client) Version(ctx context.Context) (string, error) {
 	return response.Version, nil
 }
 
+// Status returns the current state of the KES server.
+func (c *Client) Status(ctx context.Context) (State, error) {
```
Alevsk, line 313:
```suggestion
func (c *Client) Status(ctx context.Context) (*State, error) {
```

It would be better to return `*State` instead of `State`? so return will be: `return nil, err` instead of returning an empty structure

Alevsk (approved):

PR #192, opened 2022-02-09
include client IP address in audit log

aead:
With this commit the KES server adds the
client IP address to the audit events.

```
{
  "time": "2022-02-09T23:13:22+01:00",
  "request": {
    "ip": "127.0.0.1",
    "path": "/v1/log/audit/trace",
    "identity": "3ecfcdf38fcbe141ae26a1030f81e96b753365a46760ae6b578698a97c59fd22"
  },
  "response": {
    "code": 200,
    "time": 45000
  }
}
```

### DIFF
--- a/internal/http/handler.go
+++ b/internal/http/handler.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"net"
 	"net/http"
 	"path"
 	"strconv"
@@ -104,7 +105,7 @@ func enforcePolicies(config *ServerConfig, f http.HandlerFunc) http.HandlerFunc 
 // back to the client.
 func audit(logger *log.Logger, f http.HandlerFunc) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		w = &AuditResponseWriter{
+		aw := &AuditResponseWriter{
 			ResponseWriter: w,
 			Logger:         logger,
 
@@ -112,7 +113,12 @@ func audit(logger *log.Logger, f http.HandlerFunc) http.HandlerFunc {
 			Identity: auth.Identify(r),
 			Time:     time.Now(),
 		}
-		f(w, r)
+		if ip := auth.ForwardedIPFromContext(r.Context()); ip != nil {
+			aw.IP = ip
+		} else if addr, _, err := net.SplitHostPort(r.RemoteAddr); err == nil {
+			aw.IP = net.ParseIP(addr)
+		}
+		f(aw, r)
 	}
 }
 
--- a/internal/auth/proxy.go
+++ b/internal/auth/proxy.go
@@ -5,10 +5,13 @@
 package auth
 
 import (
+	"context"
 	"crypto/x509"
 	"encoding/pem"
+	"net"
 	"net/http"
 	"net/url"
+	"strings"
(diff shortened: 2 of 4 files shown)


internal/http/handler.go
```diff
@@ -104,15 +105,18 @@ func enforcePolicies(config *ServerConfig, f http.HandlerFunc) http.HandlerFunc
...
 
 			URL:      *r.URL,
 			Identity: auth.Identify(r),
 			Time:     time.Now(),
 		}
-		f(w, r)
+		if n := strings.LastIndexByte(r.RemoteAddr, ':'); n >= 0 {
+			aw.IP = net.ParseIP(r.RemoteAddr[:n])
+		}
+		f(aw, r)
```
harshavardhana, line 119: if this is behind a LB @aead r.RemoteAddr won't have the right address you need something like this to see the original client IP

```
// GetSourceIPFromHeaders retrieves the IP from the X-Forwarded-For, X-Real-IP
// and RFC7239 Forwarded headers (in that order)
func GetSourceIPFromHeaders(r *http.Request) string {
        var addr string

        if fwd := r.Header.Get(xForwardedFor); fwd != "" {
                // Only grab the first (client) address. Note that '192.168.0.1,
                // 10.1.1.1' is a valid key for X-Forwarded-For where addresses after
                // the first may represent forwarding proxies earlier in the chain.
                s := strings.Index(fwd, ", ")
                if s == -1 {
                        s = len(fwd)
                }
                addr = fwd[:s]
        } else if fwd := r.Header.Get(xRealIP); fwd != "" {
                // X-Real-IP should only contain one IP address (the client making the
                // request).
                addr = fwd
        } else if fwd := r.Header.Get(forwarded); fwd != "" {
                // match should contain at least two elements if the protocol was
                // specified in the Forwarded header. The first element will always be
                // the 'for=' capture, which we ignore. In the case of multiple IP
                // addresses (for=8.8.8.8, 8.8.4.4, 172.16.1.20 is valid) we only
                // extract the first, which should be the client IP.
                if match := forRegex.FindStringSubmatch(fwd); len(match) > 1 {
                        // IPv6 addresses in Forwarded headers are quoted-strings. We strip
                        // these quotes.
                        addr = strings.Trim(match[1], `"`)
                }
        }

        return addr
}
```

```
// GetSourceIP retrieves the IP from the request headers
// and falls back to r.RemoteAddr when necessary.
func GetSourceIP(r *http.Request) string {
        addr := GetSourceIPFromHeaders(r)
        if addr != "" {
                return addr
        }

        // Default to remote address if headers not set.
        addr, _, _ = net.SplitHostPort(r.RemoteAddr)
        return addr
}
```

harshavardhana (changes requested):

internal/http/handler.go
aead: PTAL - now the TLS proxy (KES requires that any HTTP proxy is also a TLS proxy, see: https://github.com/minio/kes/wiki/TLS-Proxy) propagates the `X-Forwarded-For` IP via the request context.

We don't care about `X-Real-IP` and `Forwarded` for now since they seem hardly used and not standardized. If there are cases that `X-Forwarded-For` is not sufficient we can revisit this.

harshavardhana (approved):

PR #196, opened 2022-02-16
remove HTTP trailers from `v1/key/list` API

aead:
This commit removes the HTTP trailer mechanism used
to signal errors. Instead, errors are now communicated
as API messages.

Further, this commit extends the listing API with additional
identity information like `CreatedAt` or `CreatedBy`.

Those fields will be used once KES can store keys itself.

### DIFF
--- a/key.go
+++ b/key.go
@@ -8,8 +8,9 @@ import (
 	"encoding"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"io"
-	"net/http"
+	"time"
 )
 
 // DEK is a data encryption key. It has a plaintext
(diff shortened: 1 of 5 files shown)


key.go
```diff
@@ -96,84 +97,90 @@ func (d *DEK) UnmarshalBinary(data []byte) error {
...
+
+// CreatedBy returns the identiy that created the current key.
+// It is a short-hand for Value().CreatedBy.
+func (i *KeyIterator) CreatedBy() Identity { return i.current.CreatedBy }
+
+// Next returns true if there is another KeyInfo.
+// It returns false if there are no more KeyInfo
+// objects or when the KeyIterator encounters an
+// error.
 func (i *KeyIterator) Next() bool {
```
klauspost, line 142: How about making it more "streaming-like"?
```suggestion
// Next returns nil if there is another KeyInfo.
// It returns io.EOF if there are no more KeyInfo
// objects.
// Any other error indicates an error was encountered,
// and there are no more entries.
func (i *KeyIterator) Next() error {
```

klauspost (approved):
Change if you like. I could see this potentially being easier to use by the caller.

PR #201, opened 2022-02-23
http: restructure API handlers

aead:
This commit refactors and restructures
the HTTP API handlers.

Before, a KES server API was implemented by
a chain of `http.HandlerFunc`s. Each handler
function had its specific task - e.g. ensure
that the client used the right HTTP method,
limit the request body to a max. size limit a.s.o.

Now, a KES server API is implemented by one
function that contains all the API logic.
For example, the `/v1/key/create` API is implemented
by the `createKey` function. The function takes care
of all aspects of the Create-Key API - e.g. checking
permissions, limiting the request body a.s.o.

This approach is a little bit more verbose but
is more readable and easier to reason about.
There is now only one place to look at when reasoning
about a specific API.

Further, this approach is more flexible since each
API can be implemented slightly differently. For example,
one API may want to modify the `AuditResponseWriter`
while all other APIs don't. Such subtle differences
between APIs can be implemented better with this new
approach.


harshavardhana:
is this a breaking change or just a restructure? @aead

aead:
@harshavardhana This only a restricting PR. It should not break any existing API.

harshavardhana (approved):

PR #202, opened 2022-03-02
http: fix `/v1/policy/assign` API

aead:
This commit fixes a bug in the policy API
that probably was introduced as rebase error.

The current `/v1/policy/assign` API uses parts
of the previous `v1/identity/assign` API such
that clients cannot assign a policy to an identity.

### DIFF
--- a/internal/http/handler.go
+++ b/internal/http/handler.go
@@ -668,23 +668,21 @@ func handleAssignPolicy(config *ServerConfig) http.HandlerFunc {
 			return
 		}
 
-		identity := kes.Identity(pathBase(r.URL.Path))
-		if identity.IsUnknown() {
-			Error(w, ErrIdentityUnknown)
-			return
-		}
-		if self := auth.Identify(r); self == identity {
-			Error(w, ErrSelfAssign)
-			return
-		}
-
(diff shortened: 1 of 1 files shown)


internal/http/handler.go
```diff
@@ -668,23 +668,21 @@ func handleAssignPolicy(config *ServerConfig) http.HandlerFunc {
 			return
 		}
 
-		identity := kes.Identity(pathBase(r.URL.Path))
-		if identity.IsUnknown() {
+		var request Request
+		if err = json.NewDecoder(r.Body).Decode(&request); err != nil {
```
Alevsk, line 672: Should it be possible for `json.NewDecoder(r.Body)`, that because a malformed `r.Body`, the function returns an `null` reference to `*Decoder`? We validate the object its not null before calling `Decode`

Alevsk (approved):

internal/http/handler.go
aead: Yes @Alevsk that is verified. When the body returns JSON that does not match the expected request schema, then the `Request` zero value will be used.

Also, the body is limited by a HTTP handler higher up the stack: https://github.com/minio/kes/blob/master/internal/http/handler.go#L74

See e.g. : https://go.dev/play/p/84asE1I6pez

PR #203, opened 2022-03-03
implement `io.WriterTo` for iterators

aead:
This commit implements the `io.WriterTo`
interface for {Key,Policy,Identity}Iterator.

With this commit it is possible to directly
pipe the output from e.g. `/v1/policy/list`
to another CLI command. For example:
```
kes policy ls | jq .
```

Now, the key iterator will parse the response
from `/v1/policy/list` while encoding the results
 exactly like received.

In particular, if e.g. there is no `CreatedBy`
identity in the server response there will no
empty `CreatedBy` in the stream passed to `jq`.

So basically, `kes policy ls | cat` gives the same
result as `curl <...> | cat`.


identity.go
```diff
@@ -110,6 +110,45 @@ func (i *IdentityIterator) Next() bool {
...
+		Identity  Identity  `json:"identity"`
+		Policy    string    `json:"policy,omitempty"`
+		CreatedAt time.Time `json:"created_at,omitempty"`
+		CreatedBy Identity  `json:"created_by,omitempty"`
+
+		Err string `json:"error,omitempty"`
+	}
+	if i.err != nil || i.closed {
+		return 0, i.err
+	}
```
klauspost, line 127: Will return `0, nil` if closed.  Maybe something like:

```suggestion
	if i.err != nil {
		return 0, i.err
	} else if i.closed != nil {
		return 0, errors.New("writeTo called after Close")
	}

```

identity.go
```diff
@@ -110,6 +110,45 @@ func (i *IdentityIterator) Next() bool {
...
+
+	cw := countWriter{W: w}
+	encoder := json.NewEncoder(&cw)
+	for {
+		var resp Response
+		if err := i.decoder.Decode(&resp); err != nil {
```
klauspost, line 133: So this is stateful? Meaning it will only return whatever hasn't been iterated over already? That would be a nice thing to have in the function docs, since that may be unexpected to some.

klauspost: Maybe just:
```
// WriteTo encodes and writes all remaining IdentityInfos from the current iterator position to w.
```

klauspost (approved):
LGTM

PR #205, opened 2022-03-11
log: print audit events on a single line

aead:
This commit fixes a terminal UI issue in the
`kes log` command. Before, the log command
would wrap the API path when it exceeds a certain
length.

Now, each log event is printed on a single line.


klauspost (approved):
LGTM

PR #206, opened 2022-03-15
key: switch to binary-encoded DEKs

aead:
This commit moves from JSON-encoded
DEKs to binary-encoded DEKs.

Previously, generated JSON-encoded
DEKs can still be decrypted. Newly,
generated DEKs are now encoded in
a binary message-pack format.

This decreases the size of DEKs
roughly about 35%.

Further, this commit adds
algorithm-specific keys. In general,
a crypto. key should be specific for
one crypto. algorithm. For example,
`AES256-GCM_SHA256`.

A key that is specific for a particular
algorithm cannot be used in other algorithms.

### DIFF
--- a/internal/key/ciphertext.go
+++ b/internal/key/ciphertext.go
@@ -1,0 +1,163 @@
+// Copyright 2022 - MinIO, Inc. All rights reserved.
+// Use of this source code is governed by the AGPLv3
+// license that can be found in the LICENSE file.
+
+package key
+
+import (
+	"encoding/json"
+
+	"github.com/minio/kes"
+	"github.com/tinylib/msgp/msgp"
+)
+
+// decodeCiphertext parses the given bytes as
+// ciphertext. If it fails to unmarshal the
+// given bytes, decodeCiphertext returns
+// ErrDecrypt.
+func decodeCiphertext(bytes []byte) (ciphertext, error) {
+	if len(bytes) == 0 {
+		return ciphertext{}, kes.ErrDecrypt
+	}
+
+	var c ciphertext
+	switch bytes[0] {
+	case 0x95: // msgp first byte
+		if err := c.UnmarshalBinary(bytes); err != nil {
+			return ciphertext{}, kes.ErrDecrypt
+		}
+	case 0x7b: // JSON first byte
+		if err := c.UnmarshalJSON(bytes); err != nil {
+			return ciphertext{}, kes.ErrDecrypt
+		}
+	default:
+		if err := c.UnmarshalBinary(bytes); err != nil {
+			if err = c.UnmarshalJSON(bytes); err != nil {
+				return ciphertext{}, kes.ErrDecrypt
+			}
+		}
+	}
+	return c, nil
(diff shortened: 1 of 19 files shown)


internal/key/ciphertext.go
```diff
@@ -0,0 +1,163 @@
...
+		return kes.ErrDecrypt
+	}
+	if len(b) != 0 {
+		return kes.ErrDecrypt
+	}
+
+	c.Algorithm = Algorithm(algorithm)
+	c.ID = id
+	c.IV = clone(iv[:]...)
+	c.Nonce = clone(nonce[:]...)
```
klauspost, line 114: No need to clone, you already have a unique copy
```suggestion
	c.IV = iv[:]
	c.Nonce = nonce[:]
```

internal/key/key.go
```diff
@@ -29,191 +30,270 @@ const (
 	Size = 256 / 8
 )
 
-// Key is a secret key for symmetric cryptography.
-type Key struct {
-	bytes []byte
-}
+// Algorithm is a cryptographic algorithm that requires
+// a cryptographic key.
+type Algorithm string
```
klauspost, line 35: I don't know how size sensitive you are, but obviously a uint8 would take less space - and you can autogen the stringer. Validation is easier.

klauspost (commented):
Driveby review

internal/key/key.go
aead: Yes, that is something I can explore. Since we use a binary format we can actually also avoid the "expensive" string.

aead: However, that change needs some additional code changes.

klauspost: If you add a MarshalText or MarshalJSON you could maybe make it backwards compatible that way?

harshavardhana:
Is there any performance advantage here @aead ?

aead:
> Is there any performance advantage here @aead ?

@harshavardhana Raw decoding is about 10x faster:
```
goos: darwin
goarch: amd64
pkg: github.com/minio/kes/internal/key
cpu: VirtualApple @ 2.50GHz
BenchmarkUnmarshalCiphertext
BenchmarkUnmarshalCiphertext/Binary
BenchmarkUnmarshalCiphertext/Binary-8         	11213253	       102.8 ns/op
BenchmarkUnmarshalCiphertext/JSON
BenchmarkUnmarshalCiphertext/JSON-8           	  733174	      1604 ns/op
```

aead:
Overall `Unwrap` is now 5x faster:
```
goos: darwin
goarch: amd64
pkg: github.com/minio/kes/internal/key
cpu: VirtualApple @ 2.50GHz
BenchmarkUnwrap
BenchmarkUnwrap/Binary
BenchmarkUnwrap/Binary-8         	 1375512	       872.5 ns/op
BenchmarkUnwrap/JSON
BenchmarkUnwrap/JSON-8           	  338954	      4352 ns/op 
```

harshavardhana (approved):

PR #213, opened 2022-03-23
add client-side functionality to detect server APIs

aead:
This commit adds the client-side functionality to
fetch the server APIs. A client can now use
```
client.APIs(context.Context) ([]API, error)
```

It returns a list of server APIs. A client may
use this to determine whether the KES server
supports a particular API without complex
version negotiation logic.

Further, a graceful clients can fetch API information
like timeouts or max. request body sizes dynamically.


klauspost (approved):
👍🏼

PR #229, opened 2022-04-27
cmd: display server endpoints listening on correctly

aead:
This commit fixes a UI bug when displaying the reachable
IP endpoints.

Before, the KES server always printed the IPv4 addresses of
all available network interfaces. This is incorrect and can
lead to confusion.

The KES server listens on the specified network interface(s).
This can be a specific network interface:
 - e.g. `192.168.188.162`
 - the pseudo address `0.0.0.0`

When listening on a specific IP / interface - e.g. `192.168.188.162`
then KES accepts connections only from that interface.
For example a request to `https://192.168.188.162:7373/version` will served
while a connection attempt to `https://127.0.0.1:7373/version` will be
refused.

In particular, listening on `:<port>` resp. `0.0.0.0:<port>` is not
the same as listening on `127.0.0.1:<port>` since the later listens
**only** on the loopback device.

Now, KES displays the endpoints / network interfaces it is listening on
correctly.

### DIFF
--- a/cmd/kes/server.go
+++ b/cmd/kes/server.go
@@ -332,7 +332,7 @@ func serverCmd(args []string) {
 	const margin = 10 // len("Endpoint: ")
 	quiet := quiet(quietFlag)
 	quiet.Print(blue.Sprint("Endpoint: "))
-	quiet.Println(bold.Sprint(alignEndpoints(margin, interfaceIP4Addrs(), port)))
+	quiet.Println(bold.Sprint(alignEndpoints(margin, listeningOnV4(ip), port)))
 	quiet.Println()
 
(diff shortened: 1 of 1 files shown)


cmd/kes/server.go
```diff
@@ -523,7 +535,7 @@ func serverAddr(addr string) (ip net.IP, port string) {
 	if ip == nil {
 		cli.Fatalf("invalid server address: %q", addr)
 	}
-	if ip.IsUnspecified() {
+	if ip.IsUnspecified() && !ip.Equal(net.IPv4(0, 0, 0, 0)) {
```
vadmeste, line 538: Can this condition be true ? 0.0.0.0.IsUnspecified() returns true already

klauspost (approved):
👍🏼

vadmeste (approved):
LGTM

harshavardhana (approved):

PR #238, opened 2022-06-15
cli: improve `kes policy show` output

aead:
This commit improves the `kes policy show`
command ouput.
```
kes policy show minio -k
Allow:
  · /v1/key/create/*
  · /v1/key/generate/*
  · /v1/key/decrypt/*
  · /v1/key/bulk/decrypt/*

Deny:
  · /v1/key/decrypt/internal-*
  · /v1/key/bulk/decrypt/internal-*

Created at: 2022-6-15 12:40:23
Created by: ee9ef42b0b9bdb65fc63702c8e8cbabe967306265388ca86c4be047b3d08097a
```

The output format remains JSON In case of STDOUT being not a terminal.
The new text output is easier to read and copy.


enclave.go
```diff
@@ -395,27 +395,41 @@ func (e *Enclave) SetPolicy(ctx context.Context, name string, policy *Policy) er
 // GetPolicy returns the policy with the given name.
 // It returns ErrPolicyNotFound if no such policy
 // exists.
-func (e *Enclave) GetPolicy(ctx context.Context, name string) (*Policy, error) {
+func (e *Enclave) GetPolicy(ctx context.Context, name string) (*Policy, *PolicyInfo, error) {
```
harshavardhana, line 398: isn't this a breaking change? @aead

aead: Yes, breaking change in the client SDK.
The issue here is that the `GetPolicy` call receives both (the policy and the policy metadata) from the server. However, it used to ignore the metadata part after parsing.

Without this change fetching the policy + metadata requires 2 SDK calls which a) not necessary since the server returns the information already and b) is error-prone since a concurrent API call may change the policy in between causing a data race...

harshavardhana: Why not keep this current API and add a new one? or for one I think `PolicyInfo` could be a member of `Policy` structure not sure how you feel about it? 

This way it keeps the API as is, however adds an additional field regarding Policy Info

```
var Policy struct {
     ......
     Info *PolicyInfo
```

Some thing like this..

aead: Finalized on:
```
type Policy struct {
	Allow []string // Set of allow patterns
	Deny  []string // Set of deny patterns

	Info PolicyInfo // Info contains metadata for the Policy.
}
```

While this allows a client to specify policy metadata that is not honoured by the server (e.g. `created_at`) it allows clients to add potential future metadata when creating a policy. Keeping the API backward compatible 👍

PR #248, opened 2022-06-30
Add client metrics to KES client

vadmeste:
HTTP connection to KES server has now two new metrics:
- Timeout error
- Other http errors

This is useful to monitor the quality of the connection to a KES server


aead:
Refer to #252 as alternative solution

aead:
Closing due to #252 got merged and implements an alternative solution.

PR #252, opened 2022-07-05
error: add `ConnError` type for network connection errors

aead:
This commit adds the `ConnError` type. A `ConnError`
is returned by a `Client` or `Enclave` to indicate
that an operation failed due to a network issue.
For example, a DNS error.

Users may check whether an operation return an
error of type `ConnError` and act appropriately.

### DIFF
--- a/retry.go
+++ b/retry.go
@@ -7,13 +7,13 @@ package kes
 import (
 	"context"
 	"errors"
-	"fmt"
 	"io"
 	"io/ioutil"
 	"math/rand"
 	"net"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 )
 
@@ -106,6 +106,11 @@ func (r *retry) Send(ctx context.Context, method string, endpoints []string, pat
 			return nil, err
 		}
 	}
+	if urlErr, ok := err.(*url.Error); ok {
+		if connErr, ok := urlErr.Err.(*ConnError); ok {
+			return nil, connErr
+		}
+	}
 	return response, err
 }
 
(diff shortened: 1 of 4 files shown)


retry.go
```diff
@@ -201,31 +206,32 @@ func (r *retry) Do(req *http.Request) (*http.Response, error) {
...
+func isNetworkError(err error) bool {
 	if err == nil { // fast path
 		return false
 	}
-	if netErr, ok := err.(net.Error); ok { // *url.Error implements net.Error
-		if netErr.Timeout() || netErr.Temporary() {
+
+	var netErr net.Error
+	if errors.As(err, &netErr) {
+		if netErr.Timeout() {
```
vadmeste, line 234: Something wrong here, you are using netErr.Timeout() which is just initialized

vadmeste (commented):
One comment.

Also one question @aead, you are only taking care of temporary networking errors because the KES client returns an error if KES server is down ?

aead:
> One comment.
> 
> Also one question @aead, you are only taking care of temporary networking errors because the KES client returns an error if KES server is down ?

Yes. If KES is down the KES client now returns:
```
kes: connection error: Get "https://127.0.0.1:7373/v1/status": dial tcp 127.0.0.1:7373: connect: connection refused
```

This means that calling code can check:
```
status, err := client.Status(ctx)
if err != nil {
   var connErr kes.ConnErr
   if errors.Is(err, &connErr) {
       // handle the situation where KES is not reachable
   }
}  
```

retry.go
vadmeste: yes, my bad

vadmeste:
```
                if urlErr, ok := netErr.(*url.Error); ok {
                        if errors.As(urlErr.Err, &netErr) {
                                return true
                        }   
                }   
```

this is what detects when a KES server is down, is that right ?

can you update isNetworkError() comment ? it seems it is only talking about transient networking errors

retry.go
vadmeste: yes, my bad

aead:
> return true

More or less yes. You cannot really detect whether the remote host is really down or just not responding / some other network issue. Updated the function comment 👍

PR #257, opened 2022-07-13
remove unnecessary space and add an example

harshavardhana:


cmd/kes/server.go
```diff
@@ -7,6 +7,7 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
```
harshavardhana, line 10: This is unrelated but had to be added due to some unforseen merge, the build was successful on both PRs however after merge the build failed.

aead (approved):

PR #264, opened 2022-08-03
init: define policy assignments in init file

aead:
This commit adds declarative policy assignments
for the stateful KES init config.

Now it is possible to pre-define identities and
there policies in the init config:
```
policy:
      minio:
        allow:
        - /v1/key/create/*
        - /v1/key/generate/*
        - /v1/key/decrypt/*
        - /v1/key/bulk/decrypt/*
        deny:
        - /v1/key/decrypt/internal-*
        - /v1/key/bulk/decrypt/internal-*
        identities:
        - c40bca7bfa2e4e92cd61f56580b9b7f2e5c356b6443ca6960df877ecce2e8016
```

During `kes init` the specified policy will be assigned to
all listed identities. This is a nice quality-of-life
improvement when deploying KES on e.g. K8S.

The `init` command will make sure that within the same enclave,
an identity can only be assigned to one policy and that the
identity is not equal to the sys admin or the enclave admin.

### DIFF
--- a/cmd/kes/init.go
+++ b/cmd/kes/init.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	tui "github.com/charmbracelet/lipgloss"
+	"github.com/minio/kes"
 	"github.com/minio/kes/internal/auth"
 	"github.com/minio/kes/internal/cli"
 	"github.com/minio/kes/internal/http"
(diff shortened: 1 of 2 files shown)


harshavardhana:
@aead this is also a time to perhaps write a kes bible - there are conceptually quite a lot of things to grasp.

I feel users have a hard time navigating the Wiki instead we should have docs/ folder to capture end to end deployment to conceptual understanding of KES.. since now the scope of KES has increased.

Praveenrajmani (approved):

Alevsk (approved):

PR #265, opened 2022-08-25
sys: switch to abstract filesystem implementation

aead:
This commits separates the logical Vault and Enclave
implementation from the persistence layer.

Now, state changes are persisted by an abstract filesystem
implementation. There are now four different filesystem
abstractions:
 - `VaultFS`
 - `KeyFS`
 - `PolicyFS`
 - `IdentityFS`

Implementations of these four filesystem abstractions build
the stateful KES persistence layer. On top of this persistence
layer, there are logical `Enclave`s managed by one logical `Vault`.

The `Vault` is responsible for creating/accessing/deleting enclaves.
Each `Enclave` has a separate `KeyFS`, `PolicyFS` and `IdentityFS`
to access and persist state. Further, the enclave implements logical
concepts, like caching and concurrent access / locking.

***

Additionally, this commit implements a new sync/locking mechanism at
the HTTP handler layer. When a client request is processed, the server
acquires one or multiple locks (i.e. one for the `Vault` and one for the
`Enclave`).

For example, when a client creates a key within the enclave `default`,
the server has to hold a read lock on the `Vault` and a write lock on
the enclave.
Since we don't modify the set of enclaves itself (just data within an
enclave) it is sufficient to hold just a read lock on the `Vault`.
However, we do modify the state of one particular enclave, and
therefore, hold a write lock on that enclave.

***

The new filesystem abstraction also paves the path for a distributed
stateful server implementation. Now, a cluster of distributed servers
can be implemented using a distributed filesystem implementation of
the four abstractions from above.


internal/sys/vault.go
```diff
@@ -6,48 +6,153 @@ package sys
...
+
+	if v.sealed {
+		return nil, kes.ErrSealed
+	}
+	if admin.IsUnknown() {
+		// TODO: disallow
+	}
+	if admin == v.admin {
+		// TODO: disallow
+	}
```
Alevsk, line 110: are these ones gonna be implemented in the future?

Alevsk (changes requested):

internal/sys/vault.go
aead: There are already. Check the latest revision.

Alevsk (approved):

PR #272, opened 2022-09-16
add `/v1/key/describe` API

aead:
This commit adds a describe key
API: `/v1/key/describe` to the
stateful/stateless server.

In addition, this commit adds a client
API implementation and CLI command for
fetching a key description.

### DIFF
--- a/cmd/kes/key.go
+++ b/cmd/kes/key.go
@@ -7,6 +7,7 @@ package main
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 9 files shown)


cmd/kes/key.go
```diff
@@ -43,6 +45,7 @@ func keyCmd(args []string) {
 	subCmds := commands{
 		"create": createKeyCmd,
 		"import": importKeyCmd,
+		"info":   describeKeyCmd,
```
Alevsk, line 48: Tested, works fine:

```
kes key info my-key-1 -k
Name        my-key-1
ID          ab4b710b5fd5eb88d23749dcbb8ed295
Created At  2022-10-04 21:24:22
Created By  cab1d7958797546d5699f577225771d6011ec12b6c6f9a449ec4b2f744d9aa08
```

Just a minor suggestion, this functionality is referred as `describe` in the code and also in the permissions, however the command is `info`, that was a little bit confused for me at the beginning since i didnt know what was the permission i needed to add to my policy, once i read the source code it was obvious I needed to use `describe` instead of `info`

Alevsk (approved):

PR #280, opened 2022-11-03
keserv: fix nil assignment and add tests

aead:
This commit fixes two nil assignment bugs
when en/decoding server config structs.
In both cases, an entry used to be assigned
to a potentially nil map causing a panic.

Further, this commit adds some basic tests
for reading / writing server configuration.


klauspost (dismissed):
LGTM

krisis (approved):

PR #282, opened 2022-11-03
use human readable memory limits

aead:
Instead of computing memory limits by
hand as `1024 * 1024` or `1<<20` use
human readable constant definitions.


vadmeste (approved):
LGTM

harshavardhana (approved):

PR #284, opened 2022-11-07
vault: use default approle engine

aead:
This commit fixes an bug when connecting
to Hashicorp Vault with AppRole credentials.

Before, ommiting the approle engine in the
config file no default engine path has used.
Now, if the approle engine in the config file
is empty, we default to `approle`.

RELEASE_TAGS BUG_FIX


krisis (approved):

klauspost (approved):
LGTM

PR #287, opened 2022-11-08
implement stateful client-side load balancing

aead:
This commit implements stateful client-side
load balancing. A `Client` and `Enclave` now
contain an internal `loadBalancer` which sends
requests to a list of endpoints.

The load balancer marks an endpoint as offline
when an endpoint fails to respond, e.g. due to
a network error or 5xx response code. Offline
endpoints no longer receive requests.

After some time period (currently 5m), the load
balancer marks offline endpoints as online to
check if it has become available. If the endpoint
still fails to respond it gets marked offline again for another 5m.

This process ensures that offline failing endpoints don't receive traffic while at the same time endpoints can become available again.

### DIFF
--- a/retry.go
+++ b/retry.go
@@ -14,6 +14,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


retry.go
```diff
@@ -75,18 +80,58 @@ type retry http.Client
...
+
+		lb.lock.Lock()
+		t, ok := lb.endpoints[nextEndpoint]
+		if ok && !t.IsZero() && t.Before(time.Now().Add(5*time.Minute)) {
+			lb.lock.Unlock()
+			continue
+		}
+		if !ok || !t.IsZero() {
+			lb.endpoints[nextEndpoint] = time.Time{}
+		}
```
klauspost, line 132: Could we change this slightly?

We set it to current time, but defer a call that will clear it if the call goes fine?

This means that only one attempt will be made on a previously bad node. If this call goes fine we resume using it for all calls.

This is just to reduce the number of calls made until we have confirmed it is fine.

klauspost (commented):
One tweak, otherwise LGTM.

retry.go
aead: Oh - I see now. A 2nd concurrent request could try the same node since it has been marked as online already. Yeah, that we can change 👍

klauspost: @vadmeste If the loop falls through with no attempts made, it will reset all connections. 

@aead - please correct me if I am wrong.

aead: > I think one goroutine doing healthcheck in background is probably fine and easier

@vadmeste I explicitly wanted to avoid a background job that needs to be monitored and/or cancelled.

retry.go
```diff
@@ -75,18 +80,58 @@ type retry http.Client
...
+retry:
 	for i := range endpoints {
 		nextEndpoint := endpoints[(i+R)%len(endpoints)]
+
+		lb.lock.Lock()
+		t, ok := lb.endpoints[nextEndpoint]
+		if ok && !t.IsZero() && t.Before(time.Now().Add(5*time.Minute)) {
+			lb.lock.Unlock()
+			continue
+		}
```
klauspost, line 129:
```suggestion
		if ok && !t.IsZero() {
			if t.Before(time.Now().Add(5 * time.Minute)) {
				lb.lock.Unlock()
				continue
			} else {
				// Reset time, so we do try this on other threads.
				// A success will reset the time and re-enabled the endpoint.
				lb.endpoints[nextEndpoint] = time.Now()
			}
		}
```

klauspost (commented):
I believe this is needed to complete the functionality

klauspost (approved):
👍🏼

PR #290, opened 2022-11-11
all: use human-readable size limits and mem.LimitReader

aead:
This commit uses human-readable size limits instead of untypted constants - i.e. `1 * mem.MiB` instead of `1 << 20`.

Further, replace `io.LimitReader` with `mem.LimitReader` to avoid type casting.


harshavardhana (approved):

harshavardhana:
PTAL at conflicts @aead

harshavardhana (approved):

klauspost (approved):
LGTM

PR #294, opened 2022-11-16
gcp: use gcp secretsmanager module instead of generic gcp go SDK

aead:
This commit changes the GCP dependency from the generic GCP Go SDK to the specific GCP secretmanager module: cloud.google.com/go/secretmanager

This module is already stable.


vadmeste (approved):
LGTM

PR #296, opened 2022-11-21
enclave: fix bug in enclave name query parameter construction

aead:
This commit fixes a bug when constructing the URL containg an enclave as query paramater.
Before, the URL API path got overwritten with the URL query parameter. For example:
```
/v1/key/create/my-key?enclacve=my-enclave
```
would have been overwritten to:

```
?enclacve=my-enclave
```


klauspost (approved):
LGTM

vadmeste (approved):
LGTM

PR #299, opened 2022-11-22
http: use human-readable memory limits

aead:


internal/http/server-key-api.go
```diff
@@ -517,7 +518,7 @@ func serverBulkDecryptKey(mux *http.ServeMux, config *ServerConfig) API {
 	const (
 		Method      = http.MethodPost
 		APIPath     = "/v1/key/bulk/decrypt/"
-		MaxBody     = 1 << 20
+		MaxBody     = int64(1 * mem.MiB)
```
klauspost, line 521: Is `mem.MiB` a typed constant since it needs conversion?

aead: It is - like `time.Millisecond`.

klauspost: Yeah, but that is needed because it isn't obvious what 1 unit refers to. 
Here it isn't really a surprise that it is a byte.

Typed values are fine as parameters and returned values, but for constants they don't really provide much value - on the contrary it just forces needless conversion.

aead: For example: https://go.dev/play/p/u4LmoB3TyP_C

harshavardhana (approved):

PR #300, opened 2022-11-23
gcp: fix bug when listing keys

aead:
This commit fixes two bugs in the list key
implementation for GCP SecretManager.

The first bug is caused by a additional `*`
appended to the `ListSecrets` API call.
This caused GCP to respond with an error like:
```
rpc error: code = NotFound desc = Project 'projects/<project-id>/*' not found.
```

Further, the GCP ListSecrets API returns full paths for each secrets. However, KES clients just know about the key name. Therefore, we have to strip the path prefix and just return the path base as key name.

### DIFF
--- a/internal/gcp/iterator.go
+++ b/internal/gcp/iterator.go
@@ -5,6 +5,8 @@
 package gcp
 
 import (
+	"path"
+
 	secretmanager "cloud.google.com/go/secretmanager/apiv1"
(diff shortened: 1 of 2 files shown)


kannappanr (approved):
LGTM

vadmeste (approved):
LGTM

PR #302, opened 2022-11-28
cmd: print KMS and endpoint on startup

aead:
This commit prints the KMS type and endpoint
on KES server startup.


klauspost (approved):
Perhaps add `Kind() string` and `Endpoint() string` to KMSConfig interface so this doesn't have to be kept up-to-date.

PR #303, opened 2022-11-30
keserv: fix region assignment in AWS SecretsManager.

aead:
This commit fixes an assignment bug in the AWS
SecretsManager config. This error leads to an
KES server error log message like:
```
aws: failed to create 'my-key': InvalidSignatureException: Credential should be scoped to a valid region.
	status code: 400
```


vadmeste (approved):
LGTM

kannappanr (approved):

PR #304, opened 2022-12-01
keserv: add version field to server config

aead:
This commit adds a version field to the
server configuration. It's purpose is to
reliably detect and parse incompatible
server config files in the future.

Currently, the only valid value for the
version field is "v1" and the field itself
is optional - for backward compatibility.
```
version: v1
address: 0.0.0.0:7373
...
```
Future versions of KES will eventually require
this field.

Based on the version field we can now distinguish
incompatible config files and parse or reject them with a proper messaging.

### DIFF
--- a/keserv/config.go
+++ b/keserv/config.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"io"
 	"os"
+	"strconv"
 	"time"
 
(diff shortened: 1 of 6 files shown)


klauspost (approved):
LGTM

PR #307, opened 2022-12-07
WIP: add key export to stateful KES server API

aead:
This commit adds a new `/v1/key/export` API
to the stateful KES server. It allows keys
to be exported by clients.

A major use case for this API is to enable a
stateless KES server to use a stateful KES
server as key store. Therefore, the stateless
KES server has to import and export keys
from/to the stateful KES server.

Furthermore, this commit adds the client-side
implementation for the `/v1/key/export` API
as well as a CLI command `kes key export`.


aead:
In favour of #308

PR #308, opened 2022-12-12
add new secrets API

aead:
This commit adds a new set of server APIs, client
APIs and CLI commands for secrets.

A secret is, similar to a crypto. key, an object
stored at the KES server. However, in contrast
to a key, a the value of a secret can be modified
and fetched from the client.

A typical use case for secrets are e.g. passwords
or API keys.

This commit adds five new server APIs:
 - `/v1/secret/create`
 - `/v1/secret/describe`
 - `/v1/secret/read`
 - `/v1/secret/delete`
 - `/v1/secret/list`

Note that there is currently no modify or edit
API endpoint. Further, a secret can only be
created if no such secret exists. Otherwise,
the KES server returns a `ErrSecretExists` error.

On the client side, this commit adds new SDK
functions (corresponding to the server APIs)
and CLI commands for managing secrets.
For example:
```
$ kes secret create my-secret "password123"

$ kes secret info my-secret

$ kes secret show my-secret

$ kes secret ls

$ kes secret rm my-secret
```
See: `kes secret --help`

***

One key use case of secrets is the ability to
use a stateful KES server as keystore for a
one or multiple stateless KES servers.

### DIFF
--- a/client.go
+++ b/client.go
@@ -499,6 +499,69 @@ func (c *Client) ListKeys(ctx context.Context, pattern string) (*KeyIterator, er
 	return enclave.ListKeys(ctx, pattern)
 }
 
+// CreateSecret creates a new secret with the given name.
+//
+// It returns ErrSecretExists if a secret with the same name
+// already exists.
+func (c *Client) CreateSecret(ctx context.Context, name string, value []byte, options *SecretOptions) error {
+	enclave := Enclave{
+		Endpoints:  c.Endpoints,
+		HTTPClient: c.HTTPClient,
+		lb:         c.lb,
+	}
+	return enclave.CreateSecret(ctx, name, value, options)
+}
+
+// DescribeSecret returns the SecretInfo for the given secret.
+//
+// It returns ErrSecretNotFound if no such secret exists.
+func (c *Client) DescribeSecret(ctx context.Context, name string) (*SecretInfo, error) {
+	enclave := Enclave{
+		Endpoints:  c.Endpoints,
+		HTTPClient: c.HTTPClient,
+		lb:         c.lb,
+	}
+	return enclave.DescribeSecret(ctx, name)
+}
+
+// ReadSecret returns the secret with the given name.
+//
+// It returns ErrSecretNotFound if no such secret exists.
+func (c *Client) ReadSecret(ctx context.Context, name string) ([]byte, *SecretInfo, error) {
+	enclave := Enclave{
+		Endpoints:  c.Endpoints,
+		HTTPClient: c.HTTPClient,
+		lb:         c.lb,
+	}
+	return enclave.ReadSecret(ctx, name)
+}
+
+// DeleteSecret deletes the secret with the given name.
+//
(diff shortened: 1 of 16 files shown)


cmd/kes/secret.go
```diff
@@ -0,0 +1,546 @@
...
+
+	enclave := newEnclave(enclaveName, insecureSkipVerify)
+	iterator, err := enclave.ListSecrets(ctx, pattern)
+	if err != nil {
+		if errors.Is(err, context.Canceled) {
+			os.Exit(1)
+		}
+		cli.Fatalf("failed to list secrets: %v", err)
+	}
+	defer iterator.Close()
```
krisis, line 497: wouldn't this `iterator.Close()` suffice?

secret.go
```diff
@@ -0,0 +1,298 @@
...
+		}
+		if err := encoder.Encode(resp); err != nil {
+			i.err = err
+			return cw.N, err
+		}
+	}
+}
+
+// Values returns up to the next n SecretInfo values. Subsequent
+// calls will yield further PolicyInfos if there are any.
```
krisis, line 251:
```suggestion
// calls will yield further SecretInfos if there are any.
```

secret.go
```diff
@@ -0,0 +1,298 @@
...
+	s.CreatedBy = v.CreatedBy
+
+	if s.ModTime.IsZero() {
+		s.ModTime = v.CreatedAt
+	}
+	return nil
+}
+
+// SecretIter iterates over a stream of SecretInfo objects.
+// Close the KeyIterator to release associated resources.
```
krisis, line 125:
```suggestion
// Close the SecretIterator to release associated resources.
```

krisis (commented):
Overall changes look good to me. I have a few minor comments/questions.

cmd/kes/secret.go
aead: We may still want to report an error when closing the iterator.
You iterate over the elements until `Next` returns false.

This may happen because there are no more elements or e.g. because we received an error.
Just relying on `defer` would silently ignore this error. The `defer` statement just ensures that we close the iterator and release resources. However, we should still check whether we the iteration was successful.

client.go
```diff
@@ -499,6 +499,69 @@ func (c *Client) ListKeys(ctx context.Context, pattern string) (*KeyIterator, er
...
+		lb:         c.lb,
+	}
+	return enclave.DeleteSecret(ctx, name)
+}
+
+// ListSecrets returns a SecretIter that iterates over all secrets
+// matching the pattern.
+//
+// The '*' pattern matches any secret. If pattern is empty the
+// SecretIter iterates over all secrets names.
```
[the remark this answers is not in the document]
aead: Yes, that is correct. As client you can specify either `*` or the empty string for listing all secrets.
However, this only works because the SDK turns the empty string into a `*`. See:
```
if pattern == "" { // The empty pattern never matches anything
	const MatchAll = "*"
	pattern = MatchAll
}
```

krisis (approved):

PR #312, opened 2022-12-19
add `KeyAlgorithm` enum type

aead:
This commit adds a enum for cryptographic
algorithms.

In general, a crypto. key is specific for
one crypto. algorithm. This commit exposes
the crypto. algorithm as enum.

Further, this commit changes the key data type
such that its binary representation is more
compact by using the `KeyAlgorithm` enum
instead of a string type.

In addition, this commit adds a version
identifier to the binary and text representation
of keys that allows easier backwards-compatible
changes to the key data type.


key.go
```diff
@@ -47,12 +48,117 @@ type PCP struct {
...
+func (a KeyAlgorithm) String() string {
+	switch a {
+	case KeyAlgorithmUndefined:
+		return "undefined"
+	case AES256_GCM_SHA256:
+		return "AES256-GCM_SHA256"
+	case XCHACHA20_POLY1305:
+		return "XCHACHA20-POLY1305"
+	default:
+		return "%" + strconv.Itoa(int(a))
```
vadmeste, line 72: I am not sure I understand why '%'

vadmeste (commented):
Few comments

key.go
aead: https://github.com/minio/kes/pull/312#discussion_r1071237275

vadmeste (dismissed):
LGTM, not tested

PR #321, opened 2023-01-20
cmd: add auto completion for bash and zsh

aead:
This commit adds command and flag auto-completion
for the `kes` command.

Auto completion can be installed via the
```
kes --auto-completion
```


vadmeste (approved):
LGTM

klauspost (approved):
LGTM (but can't test)

PR #325, opened 2023-01-23
reload certificates when receiving SIGHUP signal on unix

aead:
This commit adds support for event-based certificate reloading.

The KES server now reloads its certificate when
it receives a SIGHUP signal on unix systems.


aead:
Closed - superseded by #327

PR #328, opened 2023-02-09
add support for API keys

aead:
This commit adds API keys as secret-based
authentication mechanism.

An API key has the following format:
```
kes:v1:base64-encode(<key-type> | <key-bytes>)

key-type:  1 byte key algorithm identifier
key-bytes: value of the secret key
```

An example API key looks like this:
```
kes:v1:ACQpoGqx3rHHjT938Hfu5hVVQJHZWSqVI2Xp1KlYxFVw
```

Like TLS certificates, an API has a corresponding
identity. The identity of the example API key
above is:
```
0426fa9a04bc2756b92fbe8a97e1a1e07b53ecf04ed33da22c33e5c9faeb8cbb
```

Now, the `kes identity new` and `kes identity of`, by default, generate an API key resp. compute the identity from an API key.

***

Under the hood, an API key deterministically generates a private/public key pair - from which a CA-signed / self-signed certificate can be generated.

However, with API keys it is possible to generate
a TLS certificate, e.g. for mTLS authentication, on the fly from a single secret value.

Hence, API keys still boil down to private/public
key / TLS-based authentication. However, the simply key management when certificates don't need to be
signed by a publicly trusted CA but instead are used for internal machine-to-machine communication.

In particular, a client requires just the API key
while the server just requires the corresponding identity. Furthermore, the server does not hold any secret
(identities aren't secret information). So, we don't have to worry about compromise of API keys from the server-side.

### DIFF
--- a/apikey.go
+++ b/apikey.go
@@ -1,0 +1,190 @@
+// Copyright 2023 - MinIO, Inc. All rights reserved.
+// Use of this source code is governed by the AGPLv3
+// license that can be found in the LICENSE file.
+
+package kes
+
+import (
+	"crypto"
+	"crypto/ed25519"
+	"crypto/rand"
+	"crypto/sha256"
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/asn1"
+	"encoding/base64"
+	"encoding/hex"
+	"encoding/pem"
+	"errors"
+	"io"
+	"math/big"
+	"strings"
+	"time"
+)
+
+// APIKey is an object that can generate a private/public key pair.
+type APIKey interface {
+	// Public returns the public key that coresponds to
+	// the public key.
+	Public() crypto.PublicKey
+
+	// Private returns the private key that correspons
+	// to the public key.
+	Private() crypto.PrivateKey
+
+	// Identity returns the Identity of the public key.
+	//
+	// The identity is the cryptographic fingerprint of
+	// the raw DER-encoded public key as present in a
+	// corresponding X509 cerificate.
+	Identity() Identity
+
+	// String returns the APIKey's textual representation.
+	String() string
+}
+
+// GenerateAPIKey generates a new random API key. If rand
+// is nil, crypto/rand.Reader is used.
+func GenerateAPIKey(rand io.Reader) (APIKey, error) {
+	pub, priv, err := ed25519.GenerateKey(rand)
+	if err != nil {
+		return nil, err
+	}
+	id, err := ed25519Identity(pub)
+	if err != nil {
+		return nil, err
+	}
+	return &apiKey{
+		key:      priv,
+		identity: id,
+	}, nil
+}
+
+// ParseAPIKey parses a formatted APIKey and returns the
+// value it represents.
+func ParseAPIKey(s string) (APIKey, error) {
+	const (
+		Header      = "kes:v1:"
+		Ed25519Type = 0
+	)
+	if !strings.HasPrefix(s, Header) {
+		return nil, errors.New("kes: invalid API key: missing 'kes:v1:' prefix")
+	}
+	b, err := base64.StdEncoding.DecodeString(strings.TrimPrefix(s, Header))
+	if err != nil {
+		return nil, err
+	}
+	if len(b) != 1+ed25519.SeedSize {
+		return nil, errors.New("kes: invalid API key: invalid length")
+	}
+	if b[0] != Ed25519Type {
+		return nil, errors.New("kes: invalid API key: unsupported type")
+	}
+	key := ed25519.NewKeyFromSeed(b[1:])
+	id, err := ed25519Identity(key[ed25519.SeedSize:])
+	if err != nil {
+		return nil, err
+	}
+	return &apiKey{
+		key:      key,
+		identity: id,
+	}, nil
+}
+
+// CertificateOption is a function modifying the passed *x509.Certificate.
+type CertificateOption func(*x509.Certificate)
+
+// GenerateCertificate generates a new tls.Certificate from the APIKey's
+// private and public key. The certificate can be customized by specifying
+// one or multiple CertificateOptions.
+//
+// By default, the returned certificate is valid for 90 days.
+func GenerateCertificate(key APIKey, options ...CertificateOption) (tls.Certificate, error) {
+	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
+	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
+	if err != nil {
+		return tls.Certificate{}, err
+	}
+
+	template := x509.Certificate{
+		SerialNumber: serialNumber,
+		Subject: pkix.Name{
+			CommonName: key.Identity().String(),
+		},
+		NotBefore: time.Now().UTC(),
+		NotAfter:  time.Now().UTC().Add(90 * 24 * time.Hour),
+		KeyUsage:  x509.KeyUsageDigitalSignature,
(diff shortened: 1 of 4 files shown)


aead:
The API key of the current [`root.key`](https://github.com/minio/kes/blob/master/root.key) and [`root.cert`](https://github.com/minio/kes/blob/master/root.cert) files is:
```
kes:v1:AD9E7FSYWrMD+VjhI6q545cYT9YOyFxZb7UnjEepYDRc
```

apikey.go
```diff
@@ -0,0 +1,188 @@
+// Copyright 2019 - MinIO, Inc. All rights reserved.
```
klauspost, line 1:
```suggestion
// Copyright 2023 - MinIO, Inc. All rights reserved.
```

apikey.go
```diff
@@ -0,0 +1,188 @@
...
+		identity: id,
+	}, nil
+}
+
+// CertificateOption is a function modifying the passed *x509.Certificate.
+type CertificateOption func(*x509.Certificate)
+
+// GenerateCertificate generates a new tls.Certificate from the APIKey's
+// private and public key. The certificate can be customized by specifying
+// one or multple CertificateOptions.
```
klauspost, line 100:
```suggestion
// one or multiple CertificateOptions.
```

apikey.go
```diff
@@ -0,0 +1,188 @@
...
+		return tls.Certificate{}, err
+	}
+
+	template := x509.Certificate{
+		SerialNumber: serialNumber,
+		Subject: pkix.Name{
+			CommonName: key.Identity().String(),
+		},
+		NotBefore: time.Now().UTC(),
+		NotAfter:  time.Now().UTC().Add(90 * 24 * time.Hour),
```
klauspost, line 114: Maybe note in the function that default is 90 days. I suppose it can be altered by an option.

cmd/kes/identity.go
```diff
@@ -120,142 +112,162 @@ func newIdentityCmd(args []string) {
...
+	fmt.Fprintln(&buffer)
+	fmt.Fprintln(&buffer, "   "+bold.Render(key.Identity().String())+"\n")
+	fmt.Fprintln(&buffer, "The identity is not a secret. It can be shared with anyone.")
+	if keyPath != "" && certPath != "" {
+		fmt.Fprintln(&buffer)
+		fmt.Fprintf(&buffer, "The generated TLS private key is stored at: %s\n", keyPath)
+		fmt.Fprintf(&buffer, "The generated TLS certificate is stored at: %s\n", certPath)
+	}
+	fmt.Fprintln(&buffer)
+	fmt.Fprintln(&buffer, "It can be computed again via:")
```
klauspost, line 252:
```suggestion
	fmt.Fprintln(&buffer, "The identity be computed again via:")
```

cmd/kes/identity.go
```diff
@@ -120,142 +112,162 @@ func newIdentityCmd(args []string) {
...
+			confirm, err := term.ReadPassword(int(os.Stderr.Fd()))
+			if err != nil {
+				cli.Fatal(err)
+			}
+			fmt.Fprintln(os.Stderr)
+			if !bytes.Equal(p, confirm) {
+				cli.Fatal("passwords don't match")
+			}
+
+			block, err := x509.EncryptPEMBlock(rand.Reader, "PRIVATE KEY", privBytes, p, x509.PEMCipherAES256)
```
klauspost, line 213: We don't care about the deprecation notice?

```
// Deprecated: Legacy PEM encryption as specified in RFC 1423 is insecure by
// design. Since it does not authenticate the ciphertext, it is vulnerable to
// padding oracle attacks that can let an attacker recover the plaintext.
```


(though I guess this is also legacy for us)

apikey_test.go
```diff
@@ -0,0 +1,99 @@
+// Copyright 2019 - MinIO, Inc. All rights reserved.
```
klauspost, line 1:
```suggestion
// Copyright 2023 - MinIO, Inc. All rights reserved.
```

cmd/kes/identity.go
```diff
@@ -120,142 +112,162 @@ func newIdentityCmd(args []string) {
...
+	var buffer strings.Builder
+	fmt.Fprintln(&buffer, "Your API key:")
+	fmt.Fprintln(&buffer)
+	fmt.Fprintln(&buffer, "   "+bold.Render(key.String())+"\n")
+	fmt.Fprintln(&buffer, "This is the only time it is shown. Keep it secret and secure!")
+	fmt.Fprintln(&buffer)
+	fmt.Fprintln(&buffer, "Your Identity:")
+	fmt.Fprintln(&buffer)
+	fmt.Fprintln(&buffer, "   "+bold.Render(key.Identity().String())+"\n")
+	fmt.Fprintln(&buffer, "The identity is not a secret. It can be shared with anyone.")
```
klauspost, line 245: The concept is a bit vague. I mainly got triggered by "not a secret".

What I don't understand *what* it does.

It allows to confidentially communicate with the KES server, but it does *not* provide access to anything by itself. Is that correct?

Sorry for my ignorance, but I think this is important to explain (shortly here and longer in the docs)

cmd/kes/main.go
```diff
@@ -116,21 +116,52 @@ func main() {
...
 	if !ok {
-		cli.Fatal("no TLS client certificate. Environment variable 'KES_CLIENT_CERT' is not set")
+		cli.Fatalf("no TLS client certificate. Environment variable '%s' is not set", EnvClientCert)
 	}
 	if strings.TrimSpace(certPath) == "" {
-		cli.Fatal("no TLS client certificate. Environment variable 'KES_CLIENT_CERT' is empty")
+		cli.Fatalf("no TLS client certificate. Environment variable '%s' is empty", EnvClientCert)
 	}
 
 	keyPath, ok := os.LookupEnv("KES_CLIENT_KEY")
```
klauspost, line 159:
```suggestion
	keyPath, ok := os.LookupEnv(EnvClientKey)
```

cmd/kes/main.go
```diff
@@ -116,21 +116,52 @@ func main() {
...
+		if err != nil {
+			cli.Fatalf("invalid API key: %v", err)
+		}
+		cert, err := kes.GenerateCertificate(key)
+		if err != nil {
+			cli.Fatalf("failed to generate client certificate from API key: %v", err)
+		}
+
+		addr := DefaultServer
+		if env, ok := os.LookupEnv("KES_SERVER"); ok {
```
klauspost, line 142: Let's const it as well.

klauspost (commented):
Looks great.  Just some minor stuff.

cmd/kes/identity.go
aead: Yes, we just support PEM-encrypted private keys for compatibility with legacy systems requiring password-protected private keys.

cmd/kes/identity.go
aead: Totally agree 👍 
I hope this addition makes it more clear:
```
Your API key:

   kes:v1:ACiIOGvym58ihziHG7TgJjoAngeJCspMLN3LEhaiISSq

This is the only time it is shown. Keep it secret and secure!

Your Identity:

   4f731651f54865fe109065aaabd8aabe46d8eff1fe0ca0fa3f12c9ac0e216419

The identity is not a secret. It can be shared. Any peer
needs this identity in order to verify your API key.

The identity can be computed again via:

    kes identity of kes:v1:ACiIOGvym58ihziHG7TgJjoAngeJCspMLN3LEhaiISSq
```

klauspost (approved):
LGTM

PR #335, opened 2023-02-26
move handler logic into `internal/api`

aead:
This commit mainly restructure the existing code.
Therefore, it moves the HTTP API handlers from
`internal/http` to `internal/api`.

As part of this, the HTTP API implementation is
simplified. Now, each HTTP handler returns an `API`
```
type API struct {
	Method  string        // The HTTP method
	Path    string        // The URI API path
	MaxBody int64         // The max. body size the API accepts
	Timeout time.Duration // The duration after which an API request times out. 0 means no timeout

	// Handler implements the API.
	//
	// When invoked by the API's ServeHTTP method, the handler
	// can rely upon:
	//  - the request method matching the API's HTTP method.
	//  - the API path being a prefix of the request URL.
	//  - the request body being limited to the API's MaxBody size.
	//  - the request timing out after the duration specified for the API.
	Handler http.Handler
}

```

Each `API` contains a `Handler` that is invoked by the API instance which itself is a HTTP handler and ensures that the request uses the right HTTP method, limits the HTTP body etc.

This simplifies the actual handler implementation
and gets rid of some boilerplate code.

***

Further, the HTTP handlers now use the
`http.ResponseController` to set timeouts.


krisis (approved):
Great work!

PR #336, opened 2023-02-27
move from UBI to scratch containers

aead:
This commit replaces the UBI container image
with a multi-level docker build for a scratch
container.

This reduces the image size and avoids unneccesary bloat - which may be flagged by security scanners.

### DIFF
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -1,4 +1,10 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.7
+FROM --platform=linux/amd64 registry.access.redhat.com/ubi8/ubi-minimal:latest as build
+
+RUN microdnf update --nodocs && microdnf install ca-certificates --nodocs
+RUN  curl -s -q https://raw.githubusercontent.com/minio/kes/master/LICENSE -o LICENSE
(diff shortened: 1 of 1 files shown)


aead:
We include all labels required for certified redhat containers [Ref](https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/13/html/partner_integration/building-certified-container-images)

However, currently it is unknown whether scratch containers can be certified redhat containers. If not we may need to use `ubi-micro` (distroless) containers.

Dockerfile.release
```diff
@@ -1,4 +1,10 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.7
+FROM --platform=linux/amd64 registry.access.redhat.com/ubi8/ubi-minimal:latest as build
```
harshavardhana, line 1:
```suggestion
FROM --platform=linux/${TARGETARCH} registry.access.redhat.com/ubi8/ubi-minimal:8.7 as build
```

- We should stick to a release UBI container and not use `:latest` tag
- TargetArch allows multi-arch containers

harshavardhana (changes requested):

aead:
Closed in favor of #337

PR #339, opened 2023-03-06
expose keystore status in the `/v1/status` API

aead:
This commit exposes keystore status information
in the `/v1/status` API.

Now, a client application can check whether KES
has access to its keystore (and how big the latency is - rounded to ms). Additionally, the client
can determine whether the keystore is not available or not even reachable by KES.

### DIFF
--- a/internal/api/status.go
+++ b/internal/api/status.go
@@ -13,6 +13,7 @@ import (
 	"github.com/minio/kes/internal/audit"
 	"github.com/minio/kes/internal/auth"
 	"github.com/minio/kes/internal/sys"
+	"github.com/minio/kes/kms"
 )
 
 func status(config *RouterConfig) API {
@@ -24,15 +25,18 @@ func status(config *RouterConfig) API {
 		ContentType = "application/json"
 	)
 	type Response struct {
-		Version string        `json:"version"`
-		OS      string        `json:"os"`
-		Arch    string        `json:"arch"`
-		UpTime  time.Duration `json:"uptime"`
+		Version    string        `json:"version"`
+		OS         string        `json:"os"`
+		Arch       string        `json:"arch"`
+		UpTime     time.Duration `json:"uptime"`
+		CPUs       int           `json:"num_cpu"`
+		UsableCPUs int           `json:"num_cpu_used"`
+		HeapAlloc  uint64        `json:"mem_heap_used"`
+		StackAlloc uint64        `json:"mem_stack_used"`
 
-		CPUs       int    `json:"num_cpu"`
-		UsableCPUs int    `json:"num_cpu_used"`
-		HeapAlloc  uint64 `json:"mem_heap_used"`
-		StackAlloc uint64 `json:"mem_stack_used"`
+		KeyStoreLatency     int64 `json:"keystore_latency"` // In milliseconds
+		KeyStoreUnavailable bool  `json:"keystore_unavailable,omitempty"`
+		KeyStoreUnreachable bool  `json:"keystore_unreachable,omitempty"`
 	}
 	startTime := time.Now().UTC()
 	var handler http.HandlerFunc = func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


internal/api/status.go
```diff
@@ -24,15 +25,18 @@ func status(config *RouterConfig) API {
...
+		Arch       string        `json:"arch"`
+		UpTime     time.Duration `json:"uptime"`
+		CPUs       int           `json:"num_cpu"`
+		UsableCPUs int           `json:"num_cpu_used"`
+		HeapAlloc  uint64        `json:"mem_heap_used"`
+		StackAlloc uint64        `json:"mem_stack_used"`
+
+		KeyStoreLatency     time.Duration `json:"keystore_latency,omitempty"`
+		KeyStoreUnavailable bool          `json:"keystore_unavailable,omitempty"`
+		KeyStoreUnreachable bool          `json:"keystore_unreachable,omitempty"`
```
klauspost, line 39: Booleans shouldn't have negatives in the names, when avoidable, since it hurts readability:

```suggestion
		KeyStoreAvailable bool            `json:"keystore_available"`
		KeyStoreReachable bool            `json:"keystore_reachable"`
```

`if a && b || !KeyStoreUnavailable` is hard to logically parse. 

`KeyStoreAvailable` and `!KeyStoreAvailable` is easier to parse since neither have double negatives.

internal/api/status.go
```diff
@@ -135,16 +141,20 @@ func edgeStatus(config *EdgeRouterConfig) API {
...
```
klauspost, line 155:
```suggestion
		KeyStoreAvailable bool            `json:"keystore_available"`
		KeyStoreReachable bool            `json:"keystore_reachable"`
```

internal/api/status.go
```diff
@@ -63,6 +67,8 @@ func status(config *RouterConfig) API {
 			UsableCPUs: runtime.GOMAXPROCS(0),
 			HeapAlloc:  memStats.HeapAlloc,
 			StackAlloc: memStats.StackSys,
+
+			KeyStoreLatency: 1 * time.Millisecond, // The keystore is always available - set the min. latency.
```
klauspost, line 71: IMO we should just remove the omitempty:
```suggestion
```

internal/api/status.go
```diff
@@ -24,15 +25,18 @@ func status(config *RouterConfig) API {
...
+		Version    string        `json:"version"`
+		OS         string        `json:"os"`
+		Arch       string        `json:"arch"`
+		UpTime     time.Duration `json:"uptime"`
+		CPUs       int           `json:"num_cpu"`
+		UsableCPUs int           `json:"num_cpu_used"`
+		HeapAlloc  uint64        `json:"mem_heap_used"`
+		StackAlloc uint64        `json:"mem_stack_used"`
+
+		KeyStoreLatency     time.Duration `json:"keystore_latency,omitempty"`
```
klauspost, line 37: Not really a fan of `time.Duration` in JSON, since it is really only specific to Go. 

But I see you already have one in there. Let's all least make it 0-able:

```suggestion
		KeyStoreLatency     time.Duration `json:"keystore_latency"`
```

A more JSON-like approach would be a float64, with unit in the title:
```
		KeyStoreLatencySecs     float64 `json:"keystore_latency_secs"`
```

(milliseconds would work as well, though we might as well go with a base unit).

klauspost:
@aead Yeah. Technically all numbers are floats in JSON :D 

I was also initially going to propose milliseconds as int64. That would be fine by me as well.👍🏼 

If we have tri-states, maybe `*bool` + omitempty is probably the type we want.

aead:
> If we have tri-states, maybe *bool + omitempty is probably the type we want.

internal/api/status.go
```diff
@@ -166,7 +175,24 @@ func edgeStatus(config *EdgeRouterConfig) API {
 			UsableCPUs: runtime.GOMAXPROCS(0),
 			HeapAlloc:  memStats.HeapAlloc,
 			StackAlloc: memStats.StackSys,
-		})
+		}
+
+		state, err := config.Keys.Status(r.Context())
+		if err != nil {
+			response.KeyStoreUnavailable = true
+			_, response.KeyStoreUnreachable = kms.IsUnreachable(err)
```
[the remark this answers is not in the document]
Praveenrajmani: I guess this was used to preserve the default bool value (false)

Praveenrajmani (approved):

PR #347, opened 2023-04-17
gcp: use default GCP endpoint for status checks

aead:
This commit fixes a bug in the GCP SecretsManager
status check implementation.

Commit 64f37b1 changed the status check for a general TCP status check to a keystore specific HTTP status check.
As part of this change, the GCP implementation
only considered the SecretsManager endpoint
provided in the config file - which is usually
not set - and did not fallback to the GCP default: `secretmanager.googleapis.com:443`

This commit fixes this such that it is no longer
required to set the GCP endpoint to successfully
query the SecretsManager liveness.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -16,7 +16,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v3
       with:
-        go-version: 1.20.2
+        go-version: 1.20.3
         check-latest: true
       id: go
     - name: Check out code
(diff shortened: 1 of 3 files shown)


vadmeste:
@aead was this tested ? because it may be still an issue if we do not have https:// here

https://github.com/minio/kes/blob/master/internal/keystore/gcp/secret-manager.go#L104

harshavardhana (dismissed):

vadmeste (approved):
LGTM

PR #349, opened 2023-04-18
add git tag check to release process

aead:
This commit adds a check to the release process
to ensure that the git tag matches the commit
timestamp.

This ensures that the releases aren't tagged
incorrectly.


klauspost (approved):
LGTM

PR #350, opened 2023-04-18
edge: fix recursion bug in iterator

aead:
This commit fixes a recursion bug in the iterator
implementation.

### DIFF
--- a/edge/server-config.go
+++ b/edge/server-config.go
@@ -704,4 +704,4 @@ func (i *iter) Next() (string, bool) {
 	return "", false
 }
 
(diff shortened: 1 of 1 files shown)


klauspost (approved):
Ouch.

vadmeste (approved):
LGTM

klauspost:
@aead May I recommend adding [staticcheck](https://staticcheck.io/) to CI?

It catches these:
```
test.go:12:9: infinite recursive call (SA5007)
```

aead:
@klauspost Will add that in a separate PR - which probably will fix a series of hints 👍

PR #354, opened 2023-04-21
add readiness API

### LINKED ISSUE #353: kubernetes probes using container curl command or not requiring mTLS for http
#### What is the problem you want to solve?
I am running kes as a kubernetes container and would like to have proper liveness and readiness probes. For this purpose as far as I can see currently there are only two options:
1. Use a simple TCP check for kes on port 7373, which is likely unreliable/not comprehensive for testing functionality
2. Use a curl command inside the container like this, which uses a proper client identity for being able to check /v1/status
```
livenessProbe:
  exec:
    command: ['/bin/sh', '-ce', 'curl --connect-timeout 1 --cacert /opt/certs/ca.cert --cert /opt/certs/client.cert --key /opt/certs/client.key -s --fail -o /dev/null https://kes-server.local:7373/v1/status']
```
Kubernetes http probes do not support client certificate connections.

#### How do you want to solve it?
I'd like to ask for getting curl back into the original release image. It was there in the February image but was likely removed in commits to reduce the image size (I agree curl is a bit heavy, but has its value here IMO).

#### Additional context

1. Are there alternative solutions?
* running without probing
* using unreliable TCP checks only
* implement http probe endpoints in kes without mTLS requirement

4. Would your solution cause a major breaking API change?
No

5. Anything else that is important?

aead:
This commit adds a new API to check the readiness
of a KES edge server.

In general, liveness and readiness may be checked
by a system like Kubernetes to determine whether
a system (i.e. pod) is available and ready to serve
requests.

In general, the liveness probe should report
whether a system is not in a deadlock state.
In particular, it does not indicate whether
a system is "alive".

In contrast, the readiness probe should report
whether a system is ready to handle requests.
For example, if KES cannot reach its backend
keystore its readiness probe should fail since
it will not be able to process (some) requests.

For KES the liveness probe makes no sense at
the moment since it is not a distributed systems
where deadlocks can occur. Hence, it is not
implemented.

To make readiness really useful we have to change
the KES TLS behavior in a subtle way.

By default, KES doesn't not accept a TLS connection
attempt from a client without a TLS certificate
(or API key that translates to a certificate).
Hence, only clients with an API key or certificate
can check the KES liveness or readiness.

However, some systems (e.g. K8S) do not support
client certificates for liveness and readiness
probes. Hence, we allow the disabling authentication
for these two APIs in the `api` section in the
KES config file.

To make this actually work, we also have to tell
the Go TLS stack to allow TLS handshakes without
a certificate. Therefore, whenever at least one
API has a `skip_auth` config set to true, we also
set:
- `tls.RequestClientCert` instead of `tls.RequireAnyClientCert`
- `tls.VerifyClientCertIfGiven` instead of `tlsRequireAndVerifyClientCert`

Ref to: https://pkg.go.dev/crypto/tls#ClientAuthType

Now, the subtle behavior change is the following:
If the config file disables authentication for any
API, clients are no longer forced to send a client
certificate whenever they try to perform an API call.
Instead, the TLS handshake will always succeed.
However, for all APIs that require authentication
the HTTP API handler will reject the request with
the following error:
```
no client certificate is present
```
The difference is the protocol layer that is responsible
for rejecting clients without certificates. Before,
and also today in the default config, it's the Gop TLS
stack rejecting such connection attempts. However,
when authentication is disabled for at least one API,
the TLS stack will allow requests without a client
certificate and its no the HTTP / API layer that
has to enforce that the client sends a client certificate
when the API requires authentication.

This however, does not change the security model.
APIs that required authentication still require
authentication.

Fixes #353

### DIFF
--- a/internal/api/health.go
+++ b/internal/api/health.go
@@ -1,0 +1,55 @@
+// Copyright 2023 - MinIO, Inc. All rights reserved.
+// Use of this source code is governed by the AGPLv3
+// license that can be found in the LICENSE file.
+
+package api
+
+import (
+	"net/http"
+	"time"
+
+	"github.com/minio/kes-go"
+	"github.com/minio/kes/internal/auth"
+	"github.com/minio/kes/kv"
+)
+
+func edgeReady(config *EdgeRouterConfig) API {
+	var (
+		Method  = http.MethodGet
+		APIPath = "/v1/ready"
+		MaxBody int64
+		Timeout = 15 * time.Second
+		Verify  = true
+	)
+	if c, ok := config.APIConfig[APIPath]; ok {
+		if c.Timeout > 0 {
+			Timeout = c.Timeout
+		}
+		Verify = !c.InsecureSkipAuth
+	}
+	var handler http.HandlerFunc = func(w http.ResponseWriter, r *http.Request) {
+		if err := auth.VerifyRequest(r, config.Policies, config.Identities); Verify && err != nil {
+			Fail(w, err)
+			return
+		}
+
+		_, err := config.Keys.Status(r.Context())
+		if _, ok := kv.IsUnreachable(err); ok {
+			Fail(w, kes.NewError(http.StatusGatewayTimeout, err.Error()))
+			return
+		}
+		if err != nil {
+			Fail(w, kes.NewError(http.StatusBadGateway, err.Error()))
+			return
+		}
+		w.WriteHeader(http.StatusOK)
+	}
+	return API{
+		Method:  Method,
+		Path:    APIPath,
+		MaxBody: MaxBody,
+		Verify:  Verify,
+		Timeout: Timeout,
+		Handler: handler,
+	}
+}
--- a/cmd/kes/gateway.go
+++ b/cmd/kes/gateway.go
@@ -482,8 +482,24 @@ func newTLSConfig(config *edge.ServerConfig, auth string) (*tls.Config, error) {
 	switch strings.ToLower(auth) {
 	case "", "on":
 		clientAuth = tls.RequireAndVerifyClientCert
+		if config.API != nil {
+			for _, api := range config.API.Paths {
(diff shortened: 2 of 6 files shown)


internal/api/health.go
```diff
@@ -0,0 +1,75 @@
...
+	)
+	if c, ok := config.APIConfig[APIPath]; ok {
+		if c.Timeout > 0 {
+			Timeout = c.Timeout
+		}
+		Verify = !c.InsecureSkipAuth
+	}
+
+	var handler http.HandlerFunc = func(w http.ResponseWriter, r *http.Request) {
+		if err := auth.VerifyRequest(r, config.Policies, config.Identities); Verify && err != nil {
```
harshavardhana, line 30: Does this mean even if the policies and identities fail. The alive and ready will always return 200-OK ?

harshavardhana (changes requested):

internal/api/health.go
aead: The behavior is:
If auth is enabled (default) we check whether the client has permissions to check liveness resp. readiness.
If auth is disabled for this particular API via: https://github.com/minio/kes/blob/master/server-config.yaml#L76
any client can query the liveness resp. readiness API.

Currently, there is no concept of non-initialized IAM - like in MinIO. Either the KES config file is valid and the KES server starts with initialized IAM or it refuses to start.

harshavardhana: Yes but liveness and readiness must also ensure that stateless kes can access it's master key. 

Its not just about Kes is up or not - it is about if Kes can reach it's master key. So we need to add that.

aead: Okay, in that case we can query the KMS state in the readiness endpoint. While KES could still be able to process requests even if its backend KMS is down - e.g. offline cache - it seems reasonable to fail since at least some requests - e.g. create key - cannot be processed.

PR #358, opened 2023-05-05
keystore: improve caching mechanism

aead:
This commit replaces the existing key caching
with a new keystore caching implementation.

Therefore, this commit adds a new `internal/cache` package with provides:
 - The `cache.Cow` type, a copy-on-write concurrency-safe cache.
 - The `cache.Barrier` type, a concurrency primitive that provides mutual exclusion for sets.

The new `keystore.Cache` type is now built on top
of these two primitives.

The `cache.Cow` is a type-safe, concurrency-safe
and reader-optimized cache implementation.
It is well-suited as primitive for the `keystore.Cache` since KES requests mostly read keys. Writing a key, fetched from the KMS keystore, into the cache is a rare operation.

For many concurrent readers, a `cache.Cow` is a
better fit than a map with a `sync.Mutex`.

***

The `cache.Barrier` is a locking primitive that
prevents a subtle issue in the previous cache
implementation.

In general, the `keystore.Cache` should avoid
requesting the same key multiple times. For example, given N concurrent requests trying to access some
key `A`, the cache should only request `A` from
the key store once and serve the remaining N-1
request from the cache.
So, at any given point there should be at most
one request fetching `A` from the keystore while
the remaining N-1 requests wait until:
 - The active request succeeded to fetch `A`.
 - The active request failed. In this case the next request can try to fetch `A`.

On the flipside, blocking N-1 requests for `A`
should not block an incoming request trying to
fetch `B`.
So, a simple `sync.Mutex` is a sub-optimal solution since it serializes access to the keystore. Instead, we want to serialize access per key.

Such a serialization mechanism is implemented by
the `cache.Barrier` type. It is used very similar
to a `sync.Mutex`. Instead of `Lock` call `Lock(key)` and instead of `Unlock` call `Unlock(key)`.

Various `cache.Barrier` implementations have been
tested, including ones build using lock-free primitives. However, the current one seems to perform best w.r.t lock contention:

```
goos: darwin
goarch: arm64
pkg: github.com/minio/kes/internal/cache
          │ /Users/andreas/Desktop/1key_1000g_a.txt │  /Users/andreas/Desktop/1key_1000g_b.txt  │
          │            sec/op            │    sec/op     vs base          │
Barrier-8                    392.8µ ± 2%   498.7µ ± 68%  ~ (p=0.089 n=10)

goos: darwin
goarch: arm64
pkg: github.com/minio/kes/internal/cache
          │ /Users/andreas/Desktop/5key_1000g_a.txt │      /Users/andreas/Desktop/5key_1000g.txt       │
          │            sec/op             │     sec/op      vs base                  │
Barrier-8                     509.0µ ± 1%   11234.3µ ± 24%  +2107.33% (p=0.000 n=10)
```


klauspost (approved):
Solid

PR #359, opened 2023-05-08
remove package `kms` and `keserv`

aead:
This commit removes the two packages `kms` and
`keserv`. Both packages are no longer needed since better implementations are available via the `kv`
and `edge` package.

This cleans up a some code and removes two layers
of indirection.


shtripat (approved):
LGTM. Will try to do a sanity as well.

PR #362, opened 2023-05-10
Added Makefile

shtripat:


aead:
@shtripat Why is this needed? For local development you can simply run: `go install github.com/minio/kes/cmd/kes` or `go install ./...` inside the repo. A makefile seems overkill here.

shtripat:
> @shtripat Why is this needed? For local development you can simply run: `go install github.com/minio/kes/cmd/kes` or `go install ./...` inside the repo. A makefile seems overkill here.

@aead ony intention was to be able to build the current PR while functional tests github workflows. I understand that can be achieved with `go install` as well on ubuntu runners after having `go` setup.
Just to make it more uniform with other projects, I added this.

aead:
@shtripat If possible, I would prefer to use as few build steps as possible to keep complexity low.
As long as we can use just the `go` tooling we should not add another layer.

Also, have you looked at:
 - Commit 101f4cd80f4d60b1dd85ae6af2f9a4e1e241dab9
 - The testing package: https://github.com/minio/kes/blob/master/kestest/gateway_test.go

shtripat:
> @shtripat If possible, I would prefer to use as few build steps as possible to keep complexity low. As long as we can use just the `go` tooling we should not add another layer.
> 
> Also, have you looked at:
> 
> * Commit [101f4cd](https://github.com/minio/kes/commit/101f4cd80f4d60b1dd85ae6af2f9a4e1e241dab9)
> * The testing package: https://github.com/minio/kes/blob/master/kestest/gateway_test.go

Thanks for pointing @aead . I have seen this and understand that by default in memory store is used to verify APIs.
I understand long term goal is to have integration tests for all integrations like Vault, AWS, GCP, Azure etc.
Do you suggest to add separate similar tests here?
Also this Makefile PR I can close as not needed.

PR #363, opened 2023-05-11
Added unit tests for various key stores

shtripat:
Purpose of this change is to enable testing of various backend key stores like Azure, GCP, HashiCorp Vault etc. The tests use command like flags e.g. `aws.config`, `azure.config` to get details to connect to the key store. This is done to make the changes generic and so code remains agnostic to addition of new keystores.

Also this way we are making sure all backend key store integrations work just fine and later when integrated with KES server and client making calls throughKES server, there wont be any surprises.

### DIFF
--- a/edge_test/aws_test.go
+++ b/edge_test/aws_test.go
@@ -1,0 +1,40 @@
+package edge_test
+
+import (
+	"context"
+	"flag"
+	"os"
+	"testing"
+
+	"github.com/minio/kes/edge"
+)
+
+var awsConfigFile = flag.String("aws.config", "", "Path to a KES config file with AWS SecretsManager config")
+
+func TestAWS(t *testing.T) {
+	if *awsConfigFile == "" {
+		t.Skip("AWS tests disabled. Use -aws.config=<config file with AWS SecretManager config> to enable them")
+	}
+
+	file, err := os.Open(*awsConfigFile)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer file.Close()
+	srvrConfig, err := edge.ReadServerConfigYAML(file)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	store, err = srvrConfig.KeyStore.Connect(context.Background())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Run("create", TestCreate)
+	t.Run("set", TestSet)
+	t.Run("get", TestGet)
+	t.Run("list", TestList)
+	t.Run("delete", TestDelete)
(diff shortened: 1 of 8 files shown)


edge_test/fs_test.go
```diff
@@ -0,0 +1,138 @@
+package edge_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/minio/kes/edge"
+)
+
```
aead, line 9: Maybe it makes sense to use CLI flags to customize the tests.
```
var FSPath string

func init() {
	flag.StringVar(&FSPath, "fs.path", "", "Path used for FS tests")
}
```

Then you can run: 
```
go test -v -fs.path=<some-path>
```

edge_test/fs_test.go
```diff
@@ -0,0 +1,138 @@
...
+func TestFS_Create(t *testing.T) {
+	config := edge.FSKeyStore{
```
aead, line 11: Skip the tests if we don't have a FS config:
```
if FSPath == "" {
   t.Skip("FS tests disabled. Use -fs.path=<path> to enable them")
}
```

edge_test/aws_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+func TestAWS(t *testing.T) {
+	if *awsConfigFile == "" {
+		t.Skip("AWS tests disabled. Use -aws.config=<config file with AWS SecretManager config> to enable them")
+	}
+
+	file, err := os.Open(*awsConfigFile)
+	if err != nil {
+		t.Fatal(err)
+	}
```
aead, line 22:
```suggestion
	}
	defer file.Close()
```

aead: Same for others.

edge_test/vault_test.go
```diff
@@ -0,0 +1,40 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	store, err = srvrConfig.KeyStore.Connect(context.Background())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	os.Setenv("VAULT_SKIP_VERIFY", "true")
```
aead, line 32: We shouldn't set this. If we run the test with TLS we should set the `ca` field in the config file.

aead (changes requested):
Apart from some minor outstandings LGTM.

aead:
Also, can you squash the commits into one and add a commit message explaining the purpose of this change and why things are done in a certain way - e.g. why we use the config file CLI flag approach etc.

aead (approved):

PR #365, opened 2023-05-15
edge: handle Vault K8S auht JWTs more strictly

### LINKED ISSUE #361: Kubernetes JWT path is breaked
#### Bug describtion
Latest version stable: 2023-02-15T14-54-37Z

After this version, the kubernetes JWT field set with a file path not work well.
Kes read the file well but on the log we have this error:
```
 Error: open <WITHthe K8S token>: file name too long
```
On the log, the server try to open a file with the Token value.

#### Expected behavior

The JWT token is read only once and well used.

#### Additional context
```
    keystore:
      vault:
        endpoint: https://vault.vault.svc.cluster.local:8200
        prefix:   encryption-keys
        kubernetes:
          engine: ""
          role: kes
          jwt: /var/run/secrets/kubernetes.io/serviceaccount/token
          retry: 15s  # Duration until the server tries to re-authenticate after connection loss.
        status:
          ping: 10s
        tls:
          ca: /vault/secrets/vault.crt
```

aead:
This commit enforces that a JWT passed to the
K8S Vault engine does not require contain '/' (or
other OS path seperator).

If the JWT value contains a OS path separator we
treat it as file and read it from disk.

As a sideeffect, we require that a path to a JWT
is absolute. A relative path - e.g. `token` will
not be treated as filepath. Instead, it must be
specified as e.g. `/var/run/secrets/kubernetes.io/serviceaccount/token`.

Fixes #361

### DIFF
--- a/edge/server-config-yml.go
+++ b/edge/server-config-yml.go
@@ -423,9 +423,10 @@ func ymlToKeyStore(y *yml) (KeyStore, error) {
 				return nil, errors.New("edge: invalid vault keystore: invalid kubernetes config: no JWT specified")
 			}
 
-			// We check whether the JWT looks like a JWT (<header>.<payload>.<signature>). If not, we assume it's
-			// a path to a file containing the JWT and try to read the JWT from that file.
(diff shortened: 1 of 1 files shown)


shtripat (approved):
LGTM

PR #367, opened 2023-05-16
edge: move tests from edge_test to edge package

aead:
This commit moves the tests added by 485624f1
into the `edge` package.

Further, it converts the single tests to table-driven
tests and adds keystore cleanup to keep test runs predictable.

Tests for `List` and `Delete` are omitted because
the cleanup code first lists and then deletes entries.
So, `List` and `Delete` get tested automatically.


shtripat (approved):
LGTM. Verified

PR #368, opened 2023-05-17
Refactored `kestest` integration tests

shtripat:
This PR adds a cleanup method to make sure each test works in its entirety and there are no left outs.
Also all tests use the same common functions for scenarios like `create`, `import`, `generate`, `encrypt` and `decrypt` etc.

Different kind of tests can be executed using `go test -v -run="<test>"` command.
e.g. `go test -v -run="TestGatewayMem"` or `go test -v -fs.path=./keys -run="TestGatewayFS"`

### DIFF
--- a/kestest/gateway_test.go
+++ b/kestest/gateway_test.go
@@ -130,7 +130,7 @@ func testCreateKey(ctx context.Context, store kv.Store[string, []byte], t *testi
 	defer server.Close()
 	client := server.Client()
 
-	defer clean(ctx, client, "my-key*", t)
+	defer clean(ctx, client, t)
 
 	for i, test := range createKeyTests {
 		err := client.CreateKey(ctx, test.Name)
@@ -180,7 +180,7 @@ func testImportKey(ctx context.Context, store kv.Store[string, []byte], t *testi
 	defer server.Close()
 	client := server.Client()
 
-	defer clean(ctx, client, "my-key*", t)
+	defer clean(ctx, client, t)
 
 	for i, test := range importKeyTests {
 		err := client.ImportKey(ctx, test.Name, test.Key)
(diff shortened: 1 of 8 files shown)


kestest/gateway_test.go
```diff
@@ -633,3 +626,28 @@ func mustDecodeB64(s string) []byte {
 	}
 	return b
 }
+
+func clean(ctx context.Context, client *kes.Client, pattern string, t *testing.T) {
```
aead, line 630: You can omit the pattern here and use the match all pattern `*` in `ListKeys`.

aead (changes requested):
Minor tweaks. Also consider a descriptive commit message as suggested by the contributing doc.
Ref: https://github.com/minio/kes/blob/master/CONTRIBUTING.md#commit-message

kestest/gateway_test.go
shtripat: > You can omit the pattern here and use the match all pattern `*` in `ListKeys`.

Ok, makes sense. Will do the changes.

kestest/gateway_test.go
```diff
@@ -633,3 +626,28 @@ func mustDecodeB64(s string) []byte {
...
+
+func clean(ctx context.Context, client *kes.Client, t *testing.T) {
+	iter, err := client.ListKeys(ctx, "*")
+	if err != nil {
+		t.Fatalf("Cleanup: failed to list keys: %v", err)
+	}
+	defer iter.Close()
+
+	names := []string{}
+	keysInfo, err := iter.Values(-1)
```
aead, line 638: `iter.Values(-1)` already collects all `KeyInfo` objects into one slice. You don't need to clone this slice again. 
You can simply remove:
```
for _, keyInfo := range keysInfo {
	names = append(names, keyInfo.Name)
}
```

An just do one for loop:
```
for _, info := range keyInfos {
		if err = client.DeleteKey(ctx, info.Name); err != nil {
			t.Errorf("Cleanup: failed to delete '%s': %v", name, err)
		}
	}
```

shtripat: Done

aead (approved):

PR #386, opened 2023-08-16
vault: use `time.Ticker` instead of `time.Timer`

aead:
This commit changes the Vault re-auth. logic
to use `time.Ticker` instead of `time.Timer`.

A `time.Ticker` is better suited for periodic
tasks instead of a reseting a `time.Timer`
constantly.

Further, this commit updates the Vault SDK dep
from 1.5 to 1.9.2 and removes some deprecated API
usages.


harshavardhana (approved):

shtripat (approved):
lgtm

PR #388, opened 2023-08-17
vault: improve usage of `time.Timer`

aead:
This commit improves the usage of `time.Timer`
esp. within the Vault token renew and re-auth
logic.

It follows some recommendations from:
https://go101.org/article/memory-leaking.html

Ref: Real Memory Leaking Caused by Not Stopping
time.Ticker Values Which Are Not Used Any More

Further, this commit updates the Vault SDK dep
from 1.5 to 1.9.2 and removes some deprecated API
usages.

### DIFF
--- a/cmd/kes/gateway.go
+++ b/cmd/kes/gateway.go
@@ -12,7 +12,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"io/ioutil"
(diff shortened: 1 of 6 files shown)


harshavardhana (approved):

shtripat (approved):
lgtm

PR #393, opened 2023-08-31
vault: use custom `http.Transport`

aead:
This commit replaces the `vault.DefaultConfig`
with a custom `Config` which mainly differs from
the default by using a custom `http.Transport`.

The `vault.DefaultConfig` uses a `http.Transport`
that can leak TCP connections. Since these connections use TCP keepalive they remain idle and don't get closed causing a small but constant memory growth that can cause KES to run OOM.

### DIFF
--- a/internal/keystore/vault/vault.go
+++ b/internal/keystore/vault/vault.go
@@ -14,8 +14,10 @@ package vault
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


internal/keystore/vault/vault.go
```diff
@@ -89,8 +91,36 @@ func Connect(ctx context.Context, c *Config) (*Store, error) {
...
+	}
+	config := &vaultapi.Config{
+		Address: c.Endpoint,
+		HttpClient: &http.Client{
+			Timeout: 30 * time.Second,
+			Transport: &http.Transport{
+				Proxy:                 http.ProxyFromEnvironment,
+				DialContext:           dialer.DialContext,
+				ForceAttemptHTTP2:     true,
+				MaxIdleConns:          100,
```
harshavardhana, line 110:
```suggestion
				MaxIdleConnsPerHost:          100,
```

??

harshavardhana (changes requested):

internal/keystore/vault/vault.go
aead: The `Transport` used in this PR does not leak TCP connections. While the `Transport` used [before](https://github.com/hashicorp/go-cleanhttp/blob/v0.5.2/cleanhttp.go#L23) does cause a resource leak. Both transports differ basically on line [36](https://github.com/hashicorp/go-cleanhttp/blob/v0.5.2/cleanhttp.go#L36).

I'm not entirely sure why the resource leak occurs. However, the using `MaxIdleConns` or `MaxIdleConnsPerHost` should make no real difference since we always talk to the same host here.

PR #395, opened 2023-09-28
close keystore when reloading config

aead:
This commit ensures that the cache and keystore
get closed and any background token renewal processes are stopped when the server updates its config.

This fixes a resource leak for keystores that perform background tasks, like Vault, Gemalto, and Entrust.

### DIFF
--- a/kv/store.go
+++ b/kv/store.go
@@ -9,6 +9,7 @@ package kv
 import (
 	"context"
 	"errors"
+	"io"
 	"time"
 )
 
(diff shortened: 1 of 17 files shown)


kv/store.go
```diff
@@ -75,6 +76,8 @@ type Store[K comparable, V any] interface {
 	// List returns an Iter enumerating the stored
 	// entries.
 	List(context.Context) (Iter[K], error)
+
+	io.Closer
```
shtripat, line 80: Mostly looks good. Didn't understand the use of this.

shtripat (approved):
LGTM

PR #404, opened 2023-10-23
vault: add support for transit-encrypted K/V

aead:
This commit adds support for encrypting K/V entries with a specific transit engine key.

**Transit Engine**

The transit engine is Hashicorp Vault's en/decryption engine. Among others, it allows to send a plaintext to an encrypt API endpoint and receive a ciphertext and vice versa.
Ref: https://developer.hashicorp.com/vault/api-docs/secret/transit

Now, users can specify a transit key name in the KES config file. KES will use this key to en/decrypt its key values before storing them on the K/V backend.
However, this does, in general, not improve security since Vault encrypts all data stored on the K/V engine with internally managed keys. Users may specify a transit key if the want/have to control which key is used to encrypt the K/V data.

### DIFF
--- a/internal/keystore/vault/vault.go
+++ b/internal/keystore/vault/vault.go
@@ -83,7 +83,7 @@ func Connect(ctx context.Context, c *Config) (*Store, error) {
 			return nil, errors.New("vault: no authentication method specified")
 		}
 		if (c.AppRole.ID != "" || c.AppRole.Secret != "") && (c.K8S.JWT != "" || c.K8S.Role != "") {
-			return nil, errors.New("vault: ambigious authentication: approle and kubernetes method specified")
+			return nil, errors.New("vault: more than one authentication method specified: approle and kubernetes configuration is present")
 		}
 	}
 	if c.Transit != nil {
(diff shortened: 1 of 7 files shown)


internal/keystore/vault/vault.go
```diff
@@ -67,11 +78,18 @@ func Connect(ctx context.Context, c *Config) (*Store, error) {
...
 		return nil, fmt.Errorf("vault: invalid engine API version '%s'", c.APIVersion)
 	}
-	if (c.AppRole.ID == "" || c.AppRole.Secret == "") && (c.K8S.JWT == "" || c.K8S.Role == "") {
-		return nil, errors.New("vault: no authentication method specified")
+	if c.AppRole != nil && c.K8S != nil {
+		if (c.AppRole.ID == "" || c.AppRole.Secret == "") && (c.K8S.JWT == "" || c.K8S.Role == "") {
+			return nil, errors.New("vault: no authentication method specified")
+		}
+		if (c.AppRole.ID != "" || c.AppRole.Secret != "") && (c.K8S.JWT != "" || c.K8S.Role != "") {
+			return nil, errors.New("vault: ambigious authentication: approle and kubernetes method specified")
```
shtripat, line 86: %s/approle and kubernetes method specified/approle and kubernetes both methods specified/g

shtripat (approved):
lgtm

PR #405, opened 2023-10-24
cmd: fix regression on server address handling

aead:
This commit fixes a regression in how the `kes server` command handles the config file address.

With this commit, the server correctly uses the value from the `--addr` flag, if provided, and otherwise uses the value from the config file.


harshavardhana (approved):

kannappanr (approved):
LGTM

PR #407, opened 2023-10-25
simplify and optimize release process

aead:
This commit simplifies and optimizes the release process:
 - No longer build binaries for:
   - linux/s390x (not used)
   - linux/ppc64le (not used)
   - darwin/amd64 (apple has moved to arm64)
 - No longer build containers for:
   - linux/s390x (not used)
   - linux/ppc64le (not used)
 - Upgrade release container image to ubi-9
 - Simplify dev container build

If users really want linux/s390x or linux/ppc64le we can add these targets again. However, orgs running such infra most probably run their own registry and users can still build images for these os/arch from the docker file using `docker build -f Dockerfile .`


harshavardhana (approved):

shtripat (dismissed):

klauspost:
Why kill CodeQL?

shtripat (approved):

aead:
Based on all the reports it generated it only should false positives. Usually things like `filepath.Join` issues for "untrusted input" but args were always verified by the API handlers. On the other side, regular go linters (static-lint a.o.) seems to catch actual issues that codeql did not spot... From running it IIRC ~2y it has not been proven useful in this case. @klauspost

PR #416, opened 2023-11-09
read config file before looking up net interfaces

aead:
This commit fixes a bug in the KES server command. The server must read its config file before looking up the net addrs it listens on because the addr may not come from the command line but from the config file.


aead:
Fixed by #419

PR #417, opened 2023-11-09
kes: flush audit log events to clients

aead:
This commit fixes a bug in the KES audit log API.
The server should flush any audit log event to its clients directly. Otherwise, clients may not see
audit events since they may get buffered.


shtripat (approved):
lgtm

PR #419, opened 2023-11-09
Parse address from config if not provided with '--addr'

### LINKED ISSUE #418: 'address' is no longer parsed from config file
#### Bug describtion

I've noticed this because after updating to the latest version, kes will no longer start with the existing config.

The error message differs between the last two releases and the current state on master.

1. Releases 2023-10-27T22-05-35Z and 2023-10-24T20-26-51Z will fail with `Error: listen tcp :443: bind: permission denied` **after** printing config values that indicate that the address was parsed correctly.
2. master fails with `Error: missing port in address`.

How to reproduce:

```bash
curl -ssL https://raw.githubusercontent.com/minio/kes/master/kesconf/testdata/fs.yml > config.yml
kes identity new --key server.key --cert server.cert --ip 127.0.0.1 localhost
kes server --config ./config.yml
```

Adding the `--addr` flag solves this, whereas removing the `address` value from the config.yml file does **not**. Interestingly enough, according to the config values kes prints on startup, the address is evaluated. But when it's time to actually start the server and listen, it falls back to port 443.

Here's the config I'm downloading above:

```yaml
version: v1

address: 0.0.0.0:7373 

admin:
  identity: c84cc9b91ae2399b043da7eca616048d4b4200edf2ff418d8af3835911db945d

tls:
  key:      ./server.key  
  cert:     ./server.cert  

keystore:
  fs:
    path: "/tmp/keys" 
```

#### Expected behavior

The `address` attribute in the config should be evaluated correctly.

#### Additional context

1. What version of Go are you using (`go version`)? 1.21.4

```text
GO111MODULE='auto'
GOARCH='amd64'
GOBIN=''
GOCACHE='/home/robertl/.cache/go-build'
GOENV='/home/robertl/.config/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='amd64'
GOHOSTOS='linux'
GOINSECURE=''
GOMODCACHE='/home/robertl/go/pkg/mod'
GONOPROXY=''
GONOSUMDB=''
GOOS='linux'
GOPATH='/home/robertl/go'
GOPRIVATE=''
GOPROXY='https://proxy.golang.org,direct'
GOROOT='/usr/local/go'
GOSUMDB='sum.golang.org'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/usr/local/go/pkg/tool/linux_amd64'
GOVCS=''
GOVERSION='go1.21.4'
GCCGO='gccgo'
GOAMD64='v1'
AR='ar'
CC='gcc'
CXX='g++'
CGO_ENABLED='1'
GOMOD='/home/robertl/kes/go.mod'
GOWORK=''
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
PKG_CONFIG='pkg-config'
GOGCCFLAGS='-fPIC -m64 -pthread -Wl,--no-gc-sections -fmessage-length=0 -ffile-prefix-map=/tmp/go-build1556664813=/tmp/go-build -gno-record-gcc-switches'
```

Nope. It's trivial to reproduce.

rluetzner:
This fixes a bug where kes would try to split and evaluate the listen address, before the config file was read. This caused kes to fail if no `--addr` value was provided.

With this change, the config file is parsed before we try to determine the listen address. If the `--addr` parameter is provided, we override the value from the config as expected.

This fixes #418 .

### DIFF
--- a/cmd/kes/server.go
+++ b/cmd/kes/server.go
@@ -139,6 +139,13 @@ func startServer(addrFlag, configFlag string) error {
 		return err
 	}
 
+	rawConfig, err := kesconf.ReadFile(configFlag)
+	if err != nil {
+		return err
+	}
+	if addrFlag == "" {
+		addrFlag = rawConfig.Addr
+	}
 	host, port, err := net.SplitHostPort(addrFlag)
 	if err != nil {
 		return err
@@ -157,13 +164,6 @@ func startServer(addrFlag, configFlag string) error {
 	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
 	defer cancel()
 
-	rawConfig, err := kesconf.ReadFile(configFlag)
-	if err != nil {
-		return err
-	}
-	if addrFlag == "" {
-		addrFlag = rawConfig.Addr
-	}
 	conf, err := rawConfig.Config(ctx)
 	if err != nil {
 		return err


rluetzner:
If I remember correctly, kes previously used 7373 as it's default port. Has the default changed intentionally? If not, we should restore the default in case `addr` is neither defined as a CLI parameter or in a config file.

aead:
Dup of #416 However, LGTM 👍

aead (approved):

rluetzner:
> If I remember correctly, kes previously used 7373 as it's default port. Has the default changed intentionally? If not, we should restore the default in case `addr` is neither defined as a CLI parameter or in a config file.

@aead , have you read this comment as well? Is it expected that a listen address is defined "somewhere", either as a CLI parameter or in a config file? If no, we should add something like

```go
if (addrFlag=="") {
    addrFlag == ":7373"
}
```

after we tried the config.

aead:
I thought `kesconf` does set a default but apparently it does not. Also, it seems cleaner to read the config file as it is and not set any defaults at this layer. So, agree that the server should set the addr to `0.0.0.0:7373` if neither a CLI argument nor a config file value is present. 👍

rluetzner:
Perfect. I see you've added #420 already. 🙂

PR #421, opened 2023-11-10
fix: Avoid crash when approle is not used

vadmeste:


harshavardhana (approved):

aead:
@vadmeste Which KES version are you using to trigger this. This should've caught one layer above. However, LGTM

shtripat:
> @vadmeste Which KES version are you using to trigger this. This should've caught one layer above. However, LGTM

Agree, as with latest `kes` I see that these errors get caught while loading the config file itself. Anyway its good to have nil check added here.

vadmeste:
> @vadmeste Which KES version are you using to trigger this. This should've caught one layer above. However, LGTM

master can reproduce this. I think it is enough if you use kubernetes auth; if that's the case AppRole will be nil

PR #428, opened 2024-01-10
vault: implement authentication token renewal

aead:
This commit re-implements token renewal after it
got removed by 13cee22.

Now, the token renewal process works as following:
 - If the token is about to expire (max. 10s before expiry) we try to renew the token if renewable.
 - If the renewal fails, or the token is not renewable, KES tries to re-authenticate with the configured auth method.
 - If re-authenticating also fails, we retry after 3s.


harshavardhana:
We need a release here @aead ?

PR #414, opened 2023-10-31
kesconf: expose public API

aead:
This commit exposes kes server confguration API and provides a dev server.

Now, the `github.com/minio/kes/kesconf` package exposes a stable API for reading the KES server configuration. This supersedes the `kv` package and removes the `edge` package.

Further, the KES server provides the `--dev` flag to start a KES server in development mode with a in-memory keystore. The `--key`, `--cert` and `--auth` are deprecated. Users should use the config file to provide such config options.

PR #437, opened 2024-01-21
fix: set client CAs for mTLS auth

lu1as:
The new TLS config introduced in #414 does not set client CAs. So when enabling mTLS authentication, client requests fail with `tls: failed to verify certificate: x509: certificate signed by unknown authority`.
This PR fixes this by using the root CAs also for authenticating mTLS clients, like before:
https://github.com/minio/kes/blob/9d1b5ad6dbdd963beabfbc91eb1ca0d330d5cd3d/cmd/kes/gateway.go#L526

### DIFF
--- a/kesconf/file.go
+++ b/kesconf/file.go
@@ -150,6 +150,7 @@ func (f *File) TLSConfig() (*tls.Config, error) {
 		Certificates: []tls.Certificate{certificate},
 		NextProtos:   []string{"h2", "http/1.1"},
 		RootCAs:      rootCAs,
+		ClientCAs:    rootCAs,
 	}, nil
 }
 


aead (approved):

nicopal:
Hi, thanks for fixing this, but I think the fix is **partial** - the root **CA must be added to OS system certificates**. The error persists if KES uses a client root CA indicated by they key "ca" in the configuration file.

xstecuc:
Hi! 
At this point, do we need an identity for a client API request upon a successful mTLS auth or can the identity of the client be omitted when defining the policies? I expected it to work without identities but I keep receiving: 
`{"message":"not authorized: insufficient permissions"` 
if I don't set one.

Thank you!

PR #453, opened 2024-03-13
require a TLS client certificate by default

aead:
This commit changes the `ClientAuth` type from
`RequestClientCert` to `RequireAnyClientCert` by default.

In general, a KES server should demand a client certificate. Otherwise, a client (the HTTP/TLS stack) may choose to not send a client certificate - even if one is available. For example, the HTTP stack may try to be smart and not send a client certificate if it determines that cannot be validated since its self-signed.

Instead, the KES server's TLS should abort the handshake if the client does not send a certificate. However, in some cases we cannot enforce this. In particular, when some APIs should be accessible without TLS authentication, like `/v1/metrics`. In these cases, we have to make it optional for clients to send a certificate. However, disabling auth for some APIs is an advanced use case intended only for users who are aware of the implications.


kesconf/config.go
```diff
@@ -304,6 +304,20 @@ func ymlToServerConfig(y *ymlFile) (*File, error) {
...
+		// (the client's HTTP/TLS stack) does not send a certificate
+		// for an API that requires authentication.
+		if api.InsecureSkipAuth.Value {
+			if clientAuth == tls.RequireAnyClientCert {
+				clientAuth = tls.RequestClientCert
+			}
+			if clientAuth == tls.RequireAndVerifyClientCert {
+				clientAuth = tls.VerifyClientCertIfGiven
+			}
+		}
```
ravindk89, line 320: @aead just to confirm, how exactly would someone disable auth for an API? I don't recall this in the server config options off the top of my head.

ravindk89 (approved):
I don't see any obvious issues, but I also am not sure how to test this, so take my approval with a grain of salt.

kesconf/config.go
aead: https://github.com/minio/kes/blob/master/server-config.yaml#L59L101

harshavardhana (approved):

PR #454, opened 2024-03-18
Vault client ignore errors on return

yoks:
It looks like vault client overrides error in case if it is returned, and return unhelpful message instead.

This PR adds check if `err` is in fact `nil` before returning custom error.


shtripat (approved):
lgtm

PR #458, opened 2024-04-08
vault: improve Vault API interaction

aead:
This commit improves the Vault API interaction by
using the `vaultapi.KVv1` and `vaultapi.KVv2` types where possible. This simplifies the `List`, `Delete` and `Get` APIs.

Further, this commit changes the `List` API implementation by using HTTP GET instead of LIST. This is a workaround that prevents listing errors caused by some middleware/firewalls. Such systems may reject API calls with "non-standard" HTTP methods.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -16,7 +16,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v3
       with:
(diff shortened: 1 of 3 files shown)


shtripat (dismissed):
lgtm

shtripat (approved):

PR #430, opened 2024-01-11
fix: Upgrade azserect client for AZURE

jiuker:
Upgrade azserect client
How to test:
```shell
export ClientID=UUID
export EndPoint=url
export Secret=secret
export TenantID=tid
export ManagedIdentityClientID=micd
go test github.com/minio/kes/internal/keystore/azure
```
Once the verification is successful, I remove the redundant code
fix https://github.com/minio/kes/issues/261

PR #459, opened 2024-04-08
Upgrade to non-legacy Azure SDK

ramondeklein:
This PR upgrades from the [legacy Azure SDK](https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authorization) to the current [Azure SDK for Go](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go).


harshavardhana:
There is a PR here doing this already @ramondeklein

harshavardhana:
https://github.com/minio/kes/pull/430

ramondeklein:
Didn't see that one... Will check next time, before implementation. 

I will take the best parts of both PRs and merge it into a single PR. This PR has some advantages:
- Allows for multiple credential methods (client/secret, environment, CLI, managed identity, workload identity, ...). The existing client/secret configuration is possible, but otherwise it tries to fall back to other Azure authentication methods. Using workload or managed identity is a good alternative to client/secret, because it doesn't need any secrets (much safer).
- No runtime inspections needed to check for errors (using the `azcore.ResponseError` error type).

I do like the unit tests of the other PR, so I'll bring these over to this PR.

ramondeklein:
I started off with https://github.com/jiuker/kes/tree/upgrade_azserect_client and made some changes:
* If no explicit Azure credentials are specified (client/secret or managed identity), then it falls back to the default Azure credentials that could be stored in the environment, filesystem, ... This also allows the use of workload identifier and using Azure Keyvault during development by simply running `az login`.
* No runtime inspections needed to check for errors (using the azcore.ResponseError error type).
* Code optimization (no need for intermediate `errorResponse` type, because we already have the `status` type).
* Extended test-case.

During testing, I also found an issue that deleting secrets wasn't always handled properly. That has been fixed.

ramondeklein:
@harshavardhana I have rebased @jiuker changes to our current `master` and applied my changes in commit e59983b768645b7bcd7e53f82ecbb6c04e934562. This could supersede #430.

ramondeklein:
@aead Why does it explicitly try to load the first version? If no version is supplied, then the default Azure KeyVault behavior is to fetch the most recent version. Because we don't update secrets, there is always only one version, so the actual behaviour wouldn't change. It would make implementation a bit simpler though.

I could add that in this commit.

jiuker:
@ramondeklein fix lint plz.

internal/keystore/azure/error.go
```diff
@@ -22,18 +22,19 @@ type errorResponse struct {
...
-	limit := mem.Size(resp.ContentLength)
-	if limit < 0 || limit > MaxSize {
-		limit = MaxSize
+// transportErrToStatus converts a transport error to a Status.
+func transportErrToStatus(err error) (status, error) {
+	var rerr *azcore.ResponseError
+	if errors.As(err, &rerr) {
+		var errorResponse errorResponse
+		if rerr.RawResponse != nil {
+			json.NewDecoder(rerr.RawResponse.Body).Decode(&errorResponse)
```
harshavardhana, line 31:
```suggestion
			json.NewDecoder(rerr.RawResponse.Body).Decode(&errorResponse)
```

Lets not ignore the error from JSON parsing

harshavardhana (changes requested):
Do you have access to Azure Vault to test this?

ramondeklein:
> Do you have access to Azure Vault to test this?

I tested this in my private Azure account, but I could create instructions on how to test it in your own Azure account. As an alternative, I can sent you the credentials (via Slack) on how to test this in my account.

ramondeklein:
@harshavardhana I have added a script in `internal/keystore/azure/create-keyvault.sh` that will create an Azure key-vault in your current subscription and add the required permission to use it in the unit-test. If you don't have an Azure account, then I can sent you credentials via Slack if you would like to test for yourself.

internal/keystore/azure/error.go
```diff
@@ -22,18 +22,22 @@ type errorResponse struct {
...
-	limit := mem.Size(resp.ContentLength)
-	if limit < 0 || limit > MaxSize {
-		limit = MaxSize
+// transportErrToStatus converts a transport error to a Status.
+func transportErrToStatus(err error) (status, error) {
+	var rerr *azcore.ResponseError
+	if errors.As(err, &rerr) {
+		var errorResponse errorResponse
+		if rerr.RawResponse != nil {
+			err = json.NewDecoder(rerr.RawResponse.Body).Decode(&errorResponse)
```
jiuker, line 31: Never overwrite the input err. Just log the `Decode` error and return the inputs.

internal/keystore/azure/key-vault.go
```diff
@@ -107,24 +113,19 @@ func (s *Store) Create(ctx context.Context, name string, value []byte) error {
...
 			return fmt.Errorf("azure: failed to create '%s': failed to purge deleted secret: %s (%s)", name, stat.Message, stat.ErrorCode)
 		}
 
-		const (
-			Retry  = 7
-			Delay  = 200 * time.Millisecond
-			Jitter = 800 * time.Millisecond
-		)
-		for i := 0; i < Retry; i++ {
+		for i := 0; i < 7; i++ {
```
jiuker, line 128: If we we use `s.purgeWithRetry(ctx, name, 25)`, it's better to remove the retry.

internal/keystore/azure/key-vault.go
```diff
@@ -216,51 +217,23 @@ func (s *Store) Delete(ctx context.Context, name string) error {
 	if err != nil {
 		return fmt.Errorf("azure: failed to delete '%s': %v", name, err)
 	}
-	if stat.StatusCode != http.StatusNotFound {
```
jiuker, line 219: I think there should hold this.

jiuker (changes requested):

internal/keystore/azure/key-vault.go
ramondeklein: Even after purging, there is sometimes a bit of (unpredictable) latency before you can actually create a new secret. That's why I kept the retry mechanism here.

internal/keystore/azure/key-vault.go
ramondeklein: Not sure what you mean. The new code is resilient to cases where a non-existing key is being purged.

ramondeklein:
Processed review comment (only for reporting error for JSON deserialization issue) and rebased on the current `master` branch.

jiuker (approved):

harshavardhana (approved):

PR #461, opened 2024-05-03
cmd: support private keys in `kes identity`

aead:
This commit is part of the CLI refactoring https://github.com/minio/kes/milestone/6

This commit adds support for parsing PEM private key files and printing their private key as API key.

Now, the `kes identity` command is now able to generate API keys, print identities of API keys or certificates and convert private key files to API keys. The existing sub-commands `new`, `of`, `info` `ls` are still supported.

As of now, only private keys of type Ed25519 are supported.


cmd/kes/identity.go
```diff
@@ -45,36 +49,134 @@ Options:
...
+		fmt.Fprintln(buf)
+		fmt.Fprintln(buf, "  ", tui.NewStyle().Bold(true).Render(key.String()))
+		fmt.Fprintln(buf)
+		fmt.Fprintln(buf, "This is the only time it is shown. Keep it secret and secure!")
+		fmt.Fprintln(buf)
+		fmt.Fprintln(buf, "Your API key's identity:")
+		fmt.Fprintln(buf)
+		fmt.Fprintln(buf, "  ", tui.NewStyle().Bold(true).Render(key.Identity().String()))
+		fmt.Fprintln(buf)
+		fmt.Fprintln(buf, "The identity is not a secret an can be shared securely.")
```
shtripat, line 100:
```suggestion
		fmt.Fprintln(buf, "The identity is not a secret and can be shared securely.")
```

harshavardhana (approved):

internal/cli/term.go
```diff
@@ -0,0 +1,26 @@
...
+package cli
+
+import (
+	"os"
+
+	tui "github.com/charmbracelet/lipgloss"
+	"golang.org/x/term"
+)
+
+var isTerm = term.IsTerminal(int(os.Stdout.Fd()))
```
harshavardhana, line 14:
```suggestion
var isTerm = term.IsTerminal(int(os.Stdout.Fd())) || term.IsTerminal(int(os.Stderr.Fd()))
```

PR #465, opened 2024-06-06
cmd/kes: add support for migrating keys to minkms

aead:
This commit adds support for migrating keys to minkms via the `kes migrate` command. Migrating all keys
of a KES backend to a MinKMS server can be done as following:
```
kes migrate --from src-config.yml --server 127.0.0.1:7373 --enclave minio --api-key k1:...
```

Currently, this implementation has the following limitations:
 - The HMAC key is not migrated. This requires support from MinKMS. However, HMAC keys are not used for S3 object encryption and have been added to KES recently.
 - Ciphertexts produced by KES cannot be decrypted auto. because they lack the key version prefix (e.g. 'v1:'). Future KES servers may use ciphertexts with key versions and MinKMS may accept a ciphertext without one.

### DIFF
--- a/internal/cli/exit.go
+++ b/internal/cli/exit.go
@@ -28,3 +28,17 @@ func Exitf(format string, args ...any) {
 	fmt.Fprintln(os.Stderr, s+fmt.Sprintf(format, args...))
 	os.Exit(1)
 }
+
+// Assert calls Exit if the statement is false.
+func Assert(statement bool, args ...any) {
+	if !statement {
+		Exit(args...)
+	}
+}
+
+// Assertf calls Exitf if the statement is false.
+func Assertf(statement bool, format string, args ...any) {
+	if !statement {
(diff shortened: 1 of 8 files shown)


internal/cli/exit.go
```diff
@@ -28,3 +28,17 @@ func Exitf(format string, args ...any) {
 	fmt.Fprintln(os.Stderr, s+fmt.Sprintf(format, args...))
 	os.Exit(1)
 }
+
+// Assert calls Exit if the statement is false.
+func Assert(statement bool, args ...any) {
```
shtripat, line 33: Do we need format string for this as well?

shtripat (approved):
LGTM. One question.

internal/cli/exit.go
aead: No, we call `Exit` not `Exitf`.

PR #467, opened 2024-06-11
Fix `identity new --ip` example

ramondeklein:
Fix the `kes identity new --ip` example so it includes the `--key` and `--cert` options (required).

### DIFF
--- a/cmd/kes/identity.go
+++ b/cmd/kes/identity.go
@@ -200,7 +200,7 @@ Options:
 
 Examples:
     $ kes identity new
-    $ kes identity new --ip "192.168.0.182" --ip "10.0.0.92" localhost
+    $ kes identity new --ip "192.168.0.182" --ip "10.0.0.92" --key private.key --cert public.crt localhost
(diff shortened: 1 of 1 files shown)


cmd/kes/identity.go
```diff
@@ -249,9 +249,9 @@ func newIdentityCmd(args []string) {
 		if expiry > 0 {
 			cli.Fatalf("'--expiry' requires a private key and certificate file. Set the '--cert' and '--key' flag")
 		}
-		if len(IPs) > 0 {
-			cli.Fatalf("'--ip' requires a private key and certificate file. Set the '--cert' and '--key' flag")
-		}
+		// if len(IPs) > 0 {
```
shtripat, line 252: do we need commented lines?

shtripat (approved):
lgtm

cmd/kes/identity.go
ramondeklein: Thnx for pointing that out. I had left some debug code in there. Forgot to check my own PR 👎🏻

aead (approved):

PR #476, opened 2024-08-15
vault: clone client TLS config and headers for status checks

aead:
This commit improves the Vault status check logic. KES has to clone the Vault client since it has to clear the namespace header for status checks.

However, KES should use the same TLS client config and token configured on the client when performing such status checks. KES has to opt-in to this behavior since the Vault SDK doesn't do this by default.

### DIFF
--- a/internal/keystore/vault/vault.go
+++ b/internal/keystore/vault/vault.go
@@ -105,6 +105,8 @@ func Connect(ctx context.Context, c *Config) (*Store, error) {
 
 	config := vaultapi.DefaultConfig()
 	config.Address = c.Endpoint
+	config.CloneTLSConfig = true // Required for status checks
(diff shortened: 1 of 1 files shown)


klauspost (approved):
LGTM

internal/keystore/vault/vault.go
```diff
@@ -105,6 +105,8 @@ func Connect(ctx context.Context, c *Config) (*Store, error) {
 
 	config := vaultapi.DefaultConfig()
 	config.Address = c.Endpoint
+	config.CloneTLSConfig = true // Required for status checks
+	config.CloneToken = true     // Required for status checks
```
harshavardhana, line 109: Another ridiculous SDK

harshavardhana (approved):

PR #481, opened 2024-08-21
vault: renew token earlier

aead:
This commit changes the Vault token renewal logic to renew a token earlier than 10s before it expires. Now, if the token has a TTL > than 1m (or 30s) the token is renewed 1m (or 30s) before it expires.

This commit also adds a `context.Context` to the Vault authentication function.

### DIFF
--- a/internal/keystore/vault/client.go
+++ b/internal/keystore/vault/client.go
@@ -70,7 +70,7 @@ func (c *client) CheckStatus(ctx context.Context, delay time.Duration) {
 //
 // To renew the auth. token see: client.RenewToken(...).
 func (c *client) AuthenticateWithAppRole(login *AppRole) authFunc {
-	return func() (*vaultapi.Secret, error) {
+	return func(ctx context.Context) (*vaultapi.Secret, error) {
 		client := c.Client
 		switch {
 		case login.Namespace == "/": // Treat '/' as the root namespace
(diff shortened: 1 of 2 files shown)


internal/keystore/vault/client.go
```diff
@@ -168,9 +168,14 @@ func (c *client) RenewToken(ctx context.Context, authenticate authFunc, secret *
 		}
 
 		// We renew the token right before it expires.
-		renewIn := ttl
-		if renewIn > 10*time.Second {
-			renewIn = ttl - 10*time.Second
+		var renewIn time.Duration
```
vadmeste, line 171: I think a  better? code is to renew at 80% of the elpased time before the expiration. This can give us peace of mind

ramondeklein (approved):
We use token renewal at 80% in more places, so it would be consistent to do so. But if the current 1 minute isn't enough, then clocks differ almost a minute (or more) and that is already a problem.

internal/keystore/vault/client.go
aead: We can also do that but this means for a TTL of 1h you renew it 12min before expiry. I would be okay with that but the next thing might be that KES is renewing the token "too early"...

ramondeklein: Too early is not a problem, too often is.

vadmeste (approved):

internal/keystore/vault/client.go
aead: One might hope but I've seen some wild security theature

PR #413, opened 2023-10-31
CF CredHub KeyStore

kinjelom:
Implementation of CredHub Keystore for MinIO's KES with Mutual TLS Authentication.

Pre-release: https://github.com/kinjelom/kes/releases/tag/2023-10-31T14-25-00Z

PR #484, opened 2024-08-26
CF CredHub KeyStore

kinjelom:
Implementation of CredHub Keystore for MinIO's KES with Mutual TLS Authentication.

Related to: https://github.com/minio/kes/pull/413


kinjelom:
@harshavardhana @harshavardhana Please let me know that I won't have to wait almost a year like with PR #413, only to find out later that the interfaces to implement have changed.

PR #486, opened 2024-09-10
vault: do not reuse TCP connections

aead:
This commit disables TCP connection reuse for Vault. Apparently, TCP connections to Vault might hang if Vault gets shutdown forcefully.

The downside of this commit is that KES has to re-open a new TCP connection for every interaction with Vault. However, KES should not reach out to Vault most of the time. Hence, this change seems acceptable.


kannappanr (approved):

shtripat (approved):
lgtm

poornas (approved):

PR #488, opened 2024-10-29
vault: delay the usage of renewed auth tokens

aead:
This commit adds a delay that has to ellapse before a new Vault auth. token is used. A dist. Vault cluster may experience some replication lag. Hence, some cluster nodes already have the renewed token while others don't. If KES makes a request to a Vault node that does not have the renewed token, then the Vault node will reject the request.

By waiting a certain amount of time (e.g. 30s) KES allows the Vault cluster to sync a renewed token to all peer nodes. KES cannot know when the token will be replicated but 30s seems a reasonable value for most practical purposes.

### DIFF
--- a/internal/keystore/vault/client.go
+++ b/internal/keystore/vault/client.go
@@ -141,12 +141,16 @@ type authFunc func(context.Context) (*vaultapi.Secret, error)
 //
 //	go client.RenewToken(ctx, login, ttl)
 func (c *client) RenewToken(ctx context.Context, authenticate authFunc, secret *vaultapi.Secret) {
-	ttl, _ := secret.TokenTTL()
+	s := secret
+	ttl, _ := s.TokenTTL()
 	if ttl == 0 {
 		return // Token has no TTL. Hence, we do not need to renew it. (long-lived)
 	}
 
-	const Retry = 3 // Retry token renewal N times before re-authenticating.
+	const (
+		Retry = 3                // Retry token renewal N times before re-authenticating.
+		Delay = 30 * time.Second // Wait a certain amount of time before using a new token to account for Vault replication lag
+	)
 	for {
 		// If Vault is sealed we have to wait
 		// until it is unsealed again.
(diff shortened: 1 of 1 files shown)


internal/keystore/vault/client.go
```diff
@@ -141,12 +141,16 @@ type authFunc func(context.Context) (*vaultapi.Secret, error)
...
+	s := secret
+	ttl, _ := s.TokenTTL()
 	if ttl == 0 {
 		return // Token has no TTL. Hence, we do not need to renew it. (long-lived)
 	}
 
-	const Retry = 3 // Retry token renewal N times before re-authenticating.
+	const (
+		Retry = 3                // Retry token renewal N times before re-authenticating.
+		Delay = 30 * time.Second // Wait a certain amount of time before using a new token to account for Vault replication lag
```
harshavardhana, line 152: should this be configurable? is 30sec enough?

harshavardhana (changes requested):

internal/keystore/vault/client.go
aead: `Delay` is the time KES waits before it uses a new token received from one of the Vault nodes.
So one (actually N/2+1) Vault nodes consider this token as valid. However, due to replication lag some of the other (N/2-1) nodes may haven't received the token and cannot validate it. Hence, KES waits for 30s to allow other Vault nodes to catch up.

There is no "correct" value since Vault replication lag can be arbitrarily large. There could be a network partition between Vault nodes such that they "never" sync. However, 30s should be sufficient for all practical purposes. If you have higher replication lag, other issues arise most probably.

harshavardhana (approved):

internal/keystore/vault/client.go
```diff
@@ -155,55 +159,66 @@ func (c *client) RenewToken(ctx context.Context, authenticate authFunc, secret *
...
+			if s, _ = authenticate(ctx); s != nil {
+				ttl, _ = s.TokenTTL()
+				token, _ := s.TokenID()
+				c.SetToken(token) // SetToken is safe to call from different go routines
+			} else {
+				select {
+				case <-ctx.Done():
+					return
+				case <-time.After(3 * time.Second): // In case of an auth failure, wait 3s and retry
+				}
```
vadmeste, line 182: I think let's sleep in all cases and not only when there is an error. If someone configures a low TTL < Delay, then it can enter a loop of authentication without sleep (I would say extreme use case, but well..)

vadmeste (commented):
Neat comment, LGTM otherwise

internal/keystore/vault/client.go
aead: Yeah, right - sleeping in all cases now

PR #491, opened 2024-11-25
Re-read JWT file for every authentication

ramondeklein:
The JWT for authentication to Hashicorp Keyvault is only read during initialization or when `SIGHUP` is received. This results in issues when the SA token expires and the Keyvault client token needs to be refreshed. This PR will re-read the JWT file for every authentication.

Related to https://subnet.min.io/issues/12909.

### DIFF
--- a/kesconf/config.go
+++ b/kesconf/config.go
@@ -447,11 +447,11 @@ func ymlToKeyStore(y *ymlFile) (KeyStore, error) {
 			// We always check for '/' and the OS-specific one make cover cases where
 			// a path is specified using '/' but the underlying OS is e.g. windows.
 			if jwt := y.KeyStore.Vault.Kubernetes.JWT.Value; strings.ContainsRune(jwt, '/') || strings.ContainsRune(jwt, os.PathSeparator) {
-				b, err := os.ReadFile(y.KeyStore.Vault.Kubernetes.JWT.Value)
+				_, err := os.ReadFile(y.KeyStore.Vault.Kubernetes.JWT.Value)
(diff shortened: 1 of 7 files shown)


kesconf/config.go
```diff
@@ -447,11 +447,11 @@ func ymlToKeyStore(y *ymlFile) (KeyStore, error) {
 			// We always check for '/' and the OS-specific one make cover cases where
 			// a path is specified using '/' but the underlying OS is e.g. windows.
 			if jwt := y.KeyStore.Vault.Kubernetes.JWT.Value; strings.ContainsRune(jwt, '/') || strings.ContainsRune(jwt, os.PathSeparator) {
-				b, err := os.ReadFile(y.KeyStore.Vault.Kubernetes.JWT.Value)
+				_, err := os.ReadFile(y.KeyStore.Vault.Kubernetes.JWT.Value)
```
ramondeklein, line 450: I kept this check to ensure that the file can be read during start-up for verification only.

aead (approved):

aead:
> How often is this expected to be called worst case?

On every re-autentication to Vault, when the session token from Vault expires. Hence, not often. E.g. once an hour @klauspost

ramondeklein:
@aead I "fixed" a unit test, but it doesn't really read the JWT from the file anymore. If we do want to do that, then we need to create a mocked Vault client. I could do that, but not sure if I should...

ramondeklein:
@aead Feel free to merge (I can't do it).

PR #502, opened 2025-01-13
Verbose vault logging

ramondeklein:
This PR adds the following functionality:
- Always log Vault authentication failures.
- Enable verbose Vault HTTP logging by setting `.keystore.vault.verbose` to `true` (default is disabled).


cmd/kes/server.go
```diff
@@ -174,18 +176,26 @@ func startServer(addrFlag, configFlag string) error {
...
+	logLevel := slog.LevelInfo
+	if rawConfig.Log != nil {
+		srv.ErrLevel.Set(rawConfig.Log.ErrLevel)
+		srv.AuditLevel.Set(rawConfig.Log.AuditLevel)
+		logLevel = rawConfig.Log.LogLevel
+	}
+	if verbose {
+		logLevel = slog.LevelDebug
+	}
+	slog.SetLogLoggerLevel(logLevel)
```
ramondeklein, line 189: This needs to be set before calling `rawConfig.Config`.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/minio/kes
 
-go 1.21
+go 1.22
```
ramondeklein, line 3: Required upgrade because `slog` only supports some functions from v1.22 and later.

kesconf/file.go
```diff
@@ -298,6 +298,9 @@ type LogConfig struct {
 	// Audit determines whether the KES server logs audit events to STDOUT.
 	// It does not en/disable audit logging in general.
 	AuditLevel slog.Level
+
+	// Log level for which to report KES diagnostic messages.
+	LogLevel slog.Level
```
aead, line 303: I don't think we need another log-level field.

Currently you can set the `ErrorLevel` to:
```
off
on
debug
info
warn
error
```

It should be sufficient to set:
```
log:
  error: DEBUG
```

kesconf/config.go
```diff
@@ -354,6 +359,7 @@ func ymlToServerConfig(y *ymlFile) (*File, error) {
 		Log: &LogConfig{
 			ErrLevel:   errLevel,
 			AuditLevel: auditLevel,
+			LogLevel:   logLevel,
```
aead, line 362: Same, not needed.

aead (changes requested):
LGTM, without the additional config field

ramondeklein:
I processed the review comments. I find it kind of nasty to use `slog.Default()`, but the logger (or log-level) isn't available from the actual `Store` implementation.

ramondeklein:
@aead Not sure if we still need this. If not, then close. Otherwise please review...

Issue #515, opened 2025-04-14
AND that should likely be an OR


jub0bs:
#### Bug description

The `&&`operator on [the following line](https://github.com/minio/kes/blob/3516828509b215fe55dde6676f219f79f101aea8/internal/keystore/vault/vault.go#L187) is highly suspicious:

```go
if errors.Is(err, context.Canceled) && errors.Is(err, context.DeadlineExceeded) {
```

Shouldn't it be `||` instead?

#### Expected behavior

```go
if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
```

jub0bs:
@harshavardhana Thanks for your reply. Sorry, I don't have time to contribute to the project beyond opening this issue. Someone else should fix it.

PR #516, opened 2025-04-14
Correct condition check

shtripat:
Fixes: https://github.com/minio/kes/issues/515

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -16,7 +16,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v5
       with:
-        go-version: 1.24.0
+        go-version: 1.24.2
       id: go
     - name: Check out code
       uses: actions/checkout@v4
@@ -31,7 +31,7 @@ jobs:
       - name: "Set up Go"
         uses: actions/setup-go@v5
         with:
-          go-version: 1.24.0
+          go-version: 1.24.2
         id: go
       - name: Check out code
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):
Generally linters would complain here, we should enable those as well.

klauspost (approved):

klauspost:
@harshavardhana and @shtripat That is not a reliable lint check. Both CAN be true with wrapped errors...

```
package main

import (
	"context"
	"errors"
	"fmt"
)

func main() {
	e1 := fmt.Errorf("%w", context.DeadlineExceeded)
	e2 := fmt.Errorf("%w", context.Canceled)
	err := fmt.Errorf("e1: %w e2: %w", e1, e2)
	if errors.Is(err, context.Canceled) && errors.Is(err, context.DeadlineExceeded) {
		fmt.Println("Matched both")
	}
}
```

shtripat:
> @harshavardhana and @shtripat That is not a reliable lint check. Both CAN be true with wrapped errors...
> 
> ```go
> package main
> 
> import (
> 	"context"
> 	"errors"
> 	"fmt"
> )
> 
> func main() {
> 	err := fmt.Errorf("%w%w", context.DeadlineExceeded, context.Canceled)
> 	if errors.Is(err, context.Canceled) && errors.Is(err, context.DeadlineExceeded) {
> 		fmt.Println("Matched both")
> 	}
> }
> ```

oh, that's interesting.

aead (approved):

PR #520, opened 2025-05-05
add `version` field to crypto requests/responses

aead:
This commit adds a `Version` field to various request and response types that identifies a particular key version in a set of master keys.


shtripat (approved):
lgtm

Issue #519, opened 2025-04-30
Keys are not created in Server Startup


ramonberrutti:
#### Bug describtion
https://github.com/minio/kes/pull/111 added an option to create keys before the server startup.
We are currently using version: `2024-04-12T13-50-00Z` and tested with the main brach also.

#### Expected behavior

The key (or keys) are expected to be created before the server startup.

PR #523, opened 2025-05-23
Fix for #519

toastsandwich:
Implemented logic to define keys present in config file.

### DIFF
--- a/keystore.go
+++ b/keystore.go
@@ -19,7 +19,7 @@ import (
 	"github.com/minio/kms-go/kes"
 )
 
-// This key will be predefined in the config file, which will be stored in keystore before server starts
+// Key structure define a cryptographic key
 type Key struct {
 	Name string // Name of cryptographic key
 }
--- a/keystore.go
+++ b/keystore.go
@@ -19,6 +19,11 @@ import (
 	"github.com/minio/kms-go/kes"
 )
 
+// This key will be predefined in the config file, which will be stored in keystore before server starts
+type Key struct {
+	Name string // Name of cryptographic key
+}
+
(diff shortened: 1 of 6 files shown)


keystore.go
```diff
@@ -19,6 +19,11 @@ import (
 	"github.com/minio/kms-go/kes"
 )
 
+// Key structure define a cryptographic key
```
shtripat, line 22:
```suggestion
// Key defines a cryptographic key
```

server.go
```diff
@@ -246,6 +245,29 @@ func (s *Server) Update(conf *Config) (io.Closer, error) {
...
+				HMACKey:   hmac,
+				CreatedAt: time.Now().UTC(),
+				CreatedBy: conf.Admin,
+			}); err != nil {
+				if err != kes.ErrKeyExists {
+					return nil, err
+				}
+			}
+		}
+	}
```
shtripat, line 270: We can define a private function for this and reuse in both places

toastsandwich: You are right, we create it as a private function. Thank you @shtripat .

server.go
```diff
@@ -212,7 +211,7 @@ func (s *Server) UpdatePolicies(policies map[string]Policy) error {
 // or policies use [Server.UpdateAdmin], [Server.UpdateTLS] or
 // [Server.UpdatePolicies]. These more specific methods are usually
 // simpler to use and more efficient.
-func (s *Server) Update(conf *Config) (io.Closer, error) {
+func (s *Server) Update(ctx context.Context, conf *Config) (io.Closer, error) {
```
aead, line 214: We cannot simply change the public API here

aead (changes requested):

